PR #5, opened 2022-01-05
Automatically pin tracklist

Cephian:
Automatically pin tracklist from the `!list` command

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,7 +2,7 @@ import discord
 import asyncio
 import os
 
-from discord import Message, TextChannel, Member, VoiceClient, ClientException, ChannelType
+from discord import Message, TextChannel, Member, VoiceClient, ClientException, ChannelType, Forbidden, HTTPException
 from os import path
 from typing import List, Tuple, Optional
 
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -176,8 +176,9 @@ async def list(message: Message):
         message_to_send += f"""
 {index+1}. {author.nick or author.name} - `{filename}`"""
 
-    # Send the message
-    await message.channel.send(message_to_send)
+    # Send the message and pin it
+    list_message = await message.channel.send(message_to_send)
+    await list_message.pin()
```
anoadragon453, line 181: This should work! And while not something that'll likely occur in Busty's, could you handle the documented error codes? https://discordpy.readthedocs.io/en/latest/api.html?highlight=pin#discord.Message.pin

Someone may forgot to give the "manage messages" permission to the bot for instance.

anoadragon453 (changes requested):
LGTM other than the comment below!

Thinking about it, we should probably document which permissions the bot needs as well... I'll write up a quick issue for that.

main.py
Cephian: Sure, added handling for the pin error codes

main.py
```diff
@@ -178,7 +178,12 @@ async def list(message: Message):
 
     # Send the message and pin it
     list_message = await message.channel.send(message_to_send)
-    await list_message.pin()
+    try:
+        await list_message.pin()
+    except Forbidden:
+        print('Insufficient permission to pin tracklist. Please give me the "manage_messages" permission and try again')
```
anoadragon453, line 184: This one we could probably send this as a message to the channel instead of printing to stdout?

Though if we do send it as a message, the text should be modified to be more bustified.

anoadragon453 (approved):
Up to you if you want to send a message vs. print. LGTM otherwise!

main.py
Cephian: I think this one is better as print instead of message since it's not really a necessary feature. If say busty couldn't join the voice chat then that might be better as a message

PR #6, opened 2022-01-05
Mention users in !list and when playing their song

Cephian:
Actually use the discord @mention feature instead of just printing users' Discord nick/name in plaintext

### DIFF
--- a/main.py
+++ b/main.py
@@ -160,7 +160,7 @@ def play_next_song(e=None):
 
         # After the current song has finished, pop another off the front and continue playing
         author, filename, local_filepath = current_channel_content.pop(0)
-        await current_channel.send(f"**Playing:** {author.nick or author.name} - `{filename}`.")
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
I think this works (haven't tested)!

Alternatively, we could use the `mention` method on `User`/`Member`: https://discordpy.readthedocs.io/en/latest/api.html?highlight=mention#discord.abc.User.mention which may be more future proof.

Cephian:
> I think this works (haven't tested)!
> 
> Alternatively, we could use the `mention` method on `User`/`Member`: https://discordpy.readthedocs.io/en/latest/api.html?highlight=mention#discord.abc.User.mention which may be more future proof.

This is definitely better, I didn't even know this was in the API. I've made the change to User.mention and tested it, so I'm just merging into master now

PR #7, opened 2022-01-05
Add environment variable configuration

### LINKED ISSUE #1: Convert Discord bot token and settings to environment variables (with defaults)
Instead of hard-coding the Discord bot token, the user should pass it via an environment variable when running the program. As an example:

```sh
BUSTY_DISCORD_ACCESS_TOKEN=xxx python main.py
```

The same could be done for settings. If a specific setting isn't provided, then a sensible default value (what the busty's server uses really) should be used instead. For example:

```sh
BUSTY_DISCORD_ACCESS_TOKEN=xxx BUSTY_SECONDS_BETWEEN_SONGS=15 python main.py
```

Prefixing with `BUSTY_` ensures that the environment variable names do not collide with any other programs.

I advocate for using environment variables instead of a settings file, as:

1. It's quicker to set environment variables than copying a sample config file and the editing values.
2. It's more compatible with containers.

For cons vs. a config file:

1. It can be messier if we end up having tons of options.

Cephian:
Closes #1 by allowing configuration via environment variables.
The list of environment variables added are

1. `BUSTY_DISCORD_TOKEN` - Discord bot API token (required)
2. `BUSTY_COOLDOWN_SECS` - Number of seconds between songs (default = 10)
3. `BUSTY_ATTACHMENT_DIR` - Directory to save attachments (default = attachments)
4. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)

### DIFF
--- a/main.py
+++ b/main.py
@@ -223,4 +223,4 @@ async def scrape_channel_media(channel: TextChannel) -> List[Tuple[Member, str, 
 if 'BUSTY_DISCORD_TOKEN' in os.environ:
     client.run(os.environ['BUSTY_DISCORD_TOKEN'])
 else:
-    print('Please pass in a Discord token via the BUSTY_DISCORD_TOKEN environment variable')
+    print('Please pass in a Discord bot token via the BUSTY_DISCORD_TOKEN environment variable')
--- a/main.py
+++ b/main.py
@@ -8,7 +8,7 @@ from typing import List, Tuple, Optional
 
 ## SETTINGS
 # How many seconds to wait in-between songs
-seconds_between_songs = int(os.environ.get('BUSTY_COOLDOWN_SECS', 10))
+seconds_between_songs = int(os.environ.get("BUSTY_COOLDOWN_SECS", 10))
 # Where to save media files locally
 attachment_directory_filepath = os.environ.get("BUSTY_ATTACHMENT_DIR", "attachments")
 # The Discord role needed to perform bot commands
--- a/main.py
+++ b/main.py
@@ -8,11 +8,11 @@ from typing import List, Tuple, Optional
 
 ## SETTINGS
 # How many seconds to wait in-between songs
-seconds_between_songs = 10
+seconds_between_songs = int(os.environ.get('BUSTY_COOLDOWN_SECS', 10))
 # Where to save media files locally
-attachment_directory_filepath = "attachments"
+attachment_directory_filepath = os.environ.get("BUSTY_ATTACHMENT_DIR", "attachments")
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -8,11 +8,11 @@
 
 ## SETTINGS
 # How many seconds to wait in-between songs
-seconds_between_songs = 10
+seconds_between_songs = int(os.environ.get('BUSTY_COOLDOWN_SECS', 10))
```
anoadragon453, line 11:
```suggestion
seconds_between_songs = int(os.environ.get("BUSTY_COOLDOWN_SECS", 10))
```

For consistency with the lines below.

main.py
```diff
@@ -220,4 +220,7 @@ async def scrape_channel_media(channel: TextChannel) -> List[Tuple[Member, str,
 
 # Connect to Discord. YOUR_BOT_TOKEN_HERE must be replaced with
 # a valid Discord bot access token.
-client.run('YOUR_BOT_TOKEN_HERE')
+if 'BUSTY_DISCORD_TOKEN' in os.environ:
+    client.run(os.environ['BUSTY_DISCORD_TOKEN'])
+else:
+    print('Please pass in a Discord token via the BUSTY_DISCORD_TOKEN environment variable')
```
anoadragon453, line 226:
```suggestion
    print('Please pass in a Discord bot token via the BUSTY_DISCORD_TOKEN environment variable')
```

anoadragon453 (changes requested):

main.py
```diff
@@ -220,4 +220,7 @@ async def scrape_channel_media(channel: TextChannel) -> List[Tuple[Member, str,
 
 # Connect to Discord. YOUR_BOT_TOKEN_HERE must be replaced with
 # a valid Discord bot access token.
-client.run('YOUR_BOT_TOKEN_HERE')
+if 'BUSTY_DISCORD_TOKEN' in os.environ:
+    client.run(os.environ['BUSTY_DISCORD_TOKEN'])
+else:
+    print('Please pass in a Discord bot token via the BUSTY_DISCORD_TOKEN environment variable')
```
anoadragon453, line 226:
```suggestion
if "BUSTY_DISCORD_TOKEN" in os.environ:
    client.run(os.environ["BUSTY_DISCORD_TOKEN"])
else:
    print("Please pass in a Discord bot token via the BUSTY_DISCORD_TOKEN environment variable")
```

Oops, and here as well. Perhaps we should add a linter to CI...

anoadragon453 (approved):

Issue #3, opened 2022-01-05
Change Busty's name to the currently playing artist and song name


anoadragon453:
Such that people can easily check which song is currently playing. Otherwise they need to check the pinned message and try and keep the current or previous song in mind.

We only have 32 characters to play with, so song names may be truncated. If this happens, we should end the title with an ellipsis character: `…`

It would be fun to place an emoji in front as well. Perhaps a random emoji (excluding some not as fun ones, like the flags) that switches per artist.

anoadragon453:
Nicks are only allowed to be 32 characters or fewer (else Discord will reject the action). However it appears that Discord won't show 32 characters in all stage elements, as shown in the screenshots below:

![image](https://user-images.githubusercontent.com/1342360/148302310-43e67e7f-afba-494d-b4a8-d94d4b249bff.png)
![image](https://user-images.githubusercontent.com/1342360/148302320-44705106-a8f6-4877-9418-f2df61c80b1e.png)
![image](https://user-images.githubusercontent.com/1342360/148302331-90a8b8ea-a1e0-4434-afba-b5c77e3c57ca.png)

The longest it will show is 27, and the shortest is 19. Still, we should set Busty's nick to the longest we can in case Discord changes their UI in the future.

anoadragon453:
Done in https://github.com/anoadragon453/busty/pull/8 and https://github.com/anoadragon453/busty/pull/10.

PR #8, opened 2022-01-05
Set the bot's nick to that of the currently playing song

anoadragon453:
Partially addresses #3. This PR sets Busty's nick to that of the currently playing song. Her original nick is restored after playback has concluded. It looks like:

image

Only partially addresses, as adding an emoji to the name will be done in a separate PR (I'm thinking of pulling the set of allowed emoji from a configurable text file).

Reviewable by commit, although there's only 2.

### DIFF
--- a/main.py
+++ b/main.py
@@ -91,7 +91,7 @@ async def stop():
 
     # Restore the bot's original nick (if it exists)
     if original_bot_nickname and current_channel:
-        bot_member = current_channel.get_member(client.user.id)
+        bot_member = current_channel.guild.get_member(client.user.id)
         await bot_member.edit(nick=original_bot_nickname)
 
 
--- a/main.py
+++ b/main.py
@@ -21,6 +21,9 @@ current_channel: Optional = None
 current_channel_content: Optional[List] = None
 # The actively connected voice client
 active_voice_client: Optional[VoiceClient] = None
+# The nickname of the bot. We need to store it as it will be
+# changed while songs are being played.
+original_bot_nickname: Optional[str] = None
 
 ## STARTUP
 # This is necessary to query server members
@@ -94,6 +97,9 @@ async def play(message: Message):
         await message.channel.send("You need to be in an active voice or stage channel, sugar.")
         return
 
+    # Get a reference to the bot's Member object
+    bot_member = message.guild.get_member(client.user.id)
+
     for voice_channel in voice_channels:
         if message.author not in voice_channel.members:
             # Skip this voice channel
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -158,11 +174,21 @@ async def inner_f():
...
 
-        # After the current song has finished, pop another off the front and continue playing
+        # Pop a song off the front of the queue and play it
         author, filename, local_filepath = current_channel_content.pop(0)
         await current_channel.send(f"**Playing:** {author.mention} - `{filename}`.")
         active_voice_client.play(discord.FFmpegPCMAudio(local_filepath), after=play_next_song)
 
+        # Change the name of the bot to that of the currently playing song.
+        # This allows people to quickly see which song is currently playing.
+        new_nick = f"{author.nick or author.name} - {filename}"
```
Cephian, line 184: It might be a good idea to use the singular "ー" character instead of " - " to squeeze out two extra characters for the song name. If you think that looks too strange then feel free to leave it as-is.

Cephian (approved):

main.py
anoadragon453: So I played around with this, and using a `:` as well.

The original (`-`) with spaces:
Screenshot from 2022-01-09 13-04-34

With an emdash (`ー`) with no spaces:
Screenshot from 2022-01-09 13-07-05

With a colon: 
Screenshot from 2022-01-09 13-07-54

So it seems that Discord actually transforms dashes into emdashes in the UI automatically anyways. Also interesting was that removing the spaces surrounding the emdash only gave us one more character, even though we were eliminating two. Personally I'm not a huge fan of the emdash with no spaces; I think having spaces is more visually pleasing.

I find `:` to be equally visually pleasing, as well as saving more characters. Unfortunately, Discord seems to have a bug here where the nick is cut off short. There should be at least an ellipsis there. Here's another example of the bug:

Screenshot from 2022-01-09 13-09-48
Screenshot from 2022-01-09 13-10-01

So it was a good idea, but we're limited by what Discord's UI gives us. From the above, I'd opt for keeping the `-` with surrounding spaces.

anoadragon453: Ah! Actually, there's not a bug with the Discord UI with a colon, it's just that that is 32 characters. So we could do a colon and add an ellipsis to the end, which while wouldn't do a whole lot for the area of the UI I'm screenshotting above, it would allow people to see one more character in the stage/voice channel user listing.

Here's a comparison of ` - ` vs. `: `

image
image
image
image

Looks OK to me, and grants us one more character.

anoadragon453: Re-requesting review just to check that these changes make sense before merging.

Cephian (approved):

PR #9, opened 2022-01-05
Add a missing return upon running out of songs to play

anoadragon453:
We should exit `play_next_song` early upon running out of songs. Otherwise we'll just go ahead and try to queue up the next song, as well as say "Chilling ..." when the show is over.


Cephian (approved):
Looks good to me!

PR #10, opened 2022-01-06
Add a random emoji to the beginning of Busty's name when playing a song

anoadragon453:
Closes #8. This PR adds a random emoji to the beginning of the nick that shows the currently playing song.

It looks like:

image
image

Based on top of #8. That PR will need to be merged before this one can be.

Reviewable commit by commit. Commits for this PR start at https://github.com/anoadragon453/busty/commit/04df2f33a16879012d3a66198462c541ff58b1b6.


emoji_list.py
```diff
@@ -0,0 +1,2038 @@
+# flake8: noqa
+# Source originally taken from https://github.com/sevenc-nanashi/discord-emoji.
+# Modified to remove the UNICODE_TO_DISCORD dictionary.
+
+DISCORD_TO_UNICODE = {
```
anoadragon453, line 5: @Cephian we may want to curate this list a bit.

emoji_list.py
```diff
@@ -0,0 +1,2038 @@
+# flake8: noqa
```
anoadragon453, line 1: This came with the file, and is a comment to indicate to [flake8](https://flake8.pycqa.org/en/latest/) not to lint this file.

emoji_list.py
anoadragon453: And by a bit I mean a lot. Some of them look broken:

image

And others don't really do anything other than be a bit distracting:

image

Perhaps we should instead start with a whitelist of appropriate/fun(ny) emoji?

Cephian: Yes, a whitelist is a good idea. The reason why some emoji look broken is because it seems Discord uses your system font to render emoji in nicknames and not the Discord emoji images.

I notice since this is a Discord emoji list it follows the sorting within Discord, so along with that I propose that for now we keep all emoji in the first three sections (People, Nature, Food) and drop other ones. This would just remove all emoji from "soccer" onward from the list.

We can always hand-pick a better list later of course, or even now if you feel up to it (I don't particularly atm).

Cephian (approved):

main.py
```diff
@@ -281,6 +287,18 @@ async def scrape_channel_media(channel: TextChannel) -> List[Tuple[Member, str,
...
+
+    # Choose a random emoji
+    encoded_random_emoji = random.choice(emoji_list)
+
+    # Decode the emoji from the unicode characters
+    decoded_random_emoji = encoded_random_emoji.encode("Latin1").decode()
+
+    return decoded_random_emoji
+
+
```
Cephian, line 301: Seems like there's an extra line here? Probably want to rebase and lint before merging

anoadragon453: Ah! That's actually PEP8 convention - top-level functions and classes should be surrounded by two blank lines: https://www.python.org/dev/peps/pep-0008/#blank-lines

The linting tools will instead enforce that.

PR #13, opened 2022-01-09
Add MIT license file

### LINKED ISSUE #11: We do not have a LICENSE file
What license should we use? I'm not sure it matters all that much, but it's good practice to have one.

https://choosealicense.com/licenses/

I'd probably opt for Apache myself, as it has slightly more commercial limitations than MIT.

anoadragon453:
Fixes #11. Note that I've added both our real names into the text.


Cephian (approved):
Looks good to me

PR #15, opened 2022-01-09
Add linting CI to the repo and a lint script for local checking

anoadragon453:
This PR adds linting for continuous integration to the repo, defined by the [.github/workflows/lint.yml](https://github.com/anoadragon453/busty/blob/anoa/lint_ci/.github/workflows/lint.yml) file. It also adds a bash script at [scripts-dev/lint.sh](https://github.com/anoadragon453/busty/blob/anoa/lint_ci/scripts-dev/lint.sh), which is useful for linting your code just before you commit. This also ensures that CI will pass before creating a PR.

The linting tools that are run are:

- [isort](https://pypi.org/project/isort/) - sorts import statements
- [flake8](https://pypi.org/project/flake8/) - lints and finds mistakes
- [black](https://pypi.org/project/black/) - opinionated code formatter

I'm open to opinions on including others! Settings for each are defined in `setup.cfg`.

Note that these need to be installed prior to running `lint.sh`. The versions of each are defined in the CI workflow, and it is the essential to install the same versions locally in order to ensure that the results when running `lint.sh` locally agree with CI.

~~Based on https://github.com/anoadragon453/busty/pull/14 as it will conflict with that PR. Unfortunately, it will also conflict with other PRs. I think we should push to merge this one ASAP after #14 though, to avoid further conflicts. This PR will be rebased after #14 merges.~~ This PR has been rebased.

Best reviewed commit-by-commit.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Set up Python 3.10
       uses: actions/setup-python@v2
       with:
-        python-version: 3.10
+        python-version: "3.10"
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,6 +5,7 @@ name: Lint
 
 on:
   push:
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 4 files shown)


anoadragon453:
A succeeding run of the CI can be seen here: https://github.com/anoadragon453/busty/runs/4768016045

Cephian (approved):
I've admittedly never used any serious code linting so my review here doesn't hold all too much weight. But I see no issues with the idea of linting the project and the parts added that I do understand.

anoadragon453:
Benefits of linting mostly from my perspective are to help automatically reduce the number of things we need to catch in code review, as well as improving readability for anyone coming back to the code later on. :slightly_smiling_face:

Issue #2, opened 2022-01-05
Use rich content to increase message visibility


anoadragon453:
The current state of affairs:

![image](https://user-images.githubusercontent.com/1342360/148162284-6fa71788-827a-422b-93a4-81e5a6f20d80.png)

This message was a bit difficult to see amongst the rest of everyone's messages. Ideally we'd make use of Discord's rich message system, which provides some padding around messages, and makes them look more official. Plus we have more options for formatting.

![image](https://user-images.githubusercontent.com/1342360/148162450-9c5e69b7-fedc-4bb4-b5d8-a0b668ada1f2.png)

anoadragon453:
The `!list` command has been converted to use embeds in https://github.com/anoadragon453/busty/pull/18.

It would be nice to also use embeds for:

- [ ] The "Playing: ..." message sent before each song.
- [ ] The "Chilling for X seconds" message sent before each song.
- [ ] The "Thanks for coming" message at the end.

Cephian:
Commit #23 Resolves the "Now Playing" and "Chillin" messages, but I forgot to include the "Thanks for Coming" so I'll leave this issue open for now

anoadragon453:
I think with https://github.com/anoadragon453/busty/pull/29 merged, this is now done :tada:

PR #23, opened 2022-01-15
Revamp "Now Playing" and "Chillin" messages

Cephian:
Resolves Issue #2.

Make the "Now Playing" and waiting messages rich text embed for better visibility. Also add a number of other features like

- Direct file jumping (like in !list)
- Direct message jumping (like in !list)
- Re-posting text sent with original submission
image

Also includes a slight internal restructure. The return tuple of the `scrape_channel_media` function was getting bloated, so it now keeps direct references to the original message and attachment from which most previously included fields can be derived.

### DIFF
--- a/main.py
+++ b/main.py
@@ -6,11 +6,11 @@ from typing import List, Optional, Tuple
 
 import discord
 from discord import (
+    Attachment,
     ChannelType,
     ClientException,
     Forbidden,
     HTTPException,
-    Member,
     Message,
     NotFound,
     TextChannel,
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -274,9 +296,9 @@ async def command_list(message: Message):
 
 async def scrape_channel_media(
     channel: TextChannel,
-) -> List[Tuple[Member, str, str, str]]:
+) -> List[Tuple[Message, Attachment, str]]:
     # A list of (uploader, filename, local filepath, message jump url)
```
anoadragon453, line 300: This comment needs an update.

```suggestion
    # A list of (original message, message attachment, local filepath)
```

main.py
```diff
@@ -195,27 +195,51 @@ async def inner_f():
 
         # Wait some time between songs
         if seconds_between_songs:
-            await current_channel.send(
-                f"Chillin' for {seconds_between_songs} seconds..."
+            embed_title = "Currently Chillin'"
+            embed_content = "Waiting for {} second{}...".format(
+                seconds_between_songs, "s" if seconds_between_songs != 1 else ""
             )
+            embed = discord.Embed(title=embed_title, description=embed_content)
```
anoadragon453, line 202: Just checking it was intentional to not include a color for this embed.

Personally I think this one doesn't need a color or need to be too loud, and it's just supposed to be a short transitional message while people get their last words in.

anoadragon453 (approved):
One minor thing - looks great though!

main.py
Cephian: Yup! it was intentional

PR #24, opened 2022-01-15
Remove some non-fun entries from the list of possible emoji

anoadragon453:
Removes the following entries from the emoji pool, as IMO they'd be a bit unfortunate if they displayed over/next to a song:

image

On the fence about:

image

though I'm leaning towards keeping it as it's pretty funny.


emoji_list.py
```diff
@@ -72,24 +67,19 @@
...
     "hugging": "\xf0\x9f\xa4\x97",
     "hugging_face": "\xf0\x9f\xa4\x97",
     "thinking": "\xf0\x9f\xa4\x94",
     "thinking_face": "\xf0\x9f\xa4\x94",
     "face_with_hand_over_mouth": "\xf0\x9f\xa4\xad",
-    "yawning_face": "\xf0\x9f\xa5\xb1",
     "shushing_face": "\xf0\x9f\xa4\xab",
     "lying_face": "\xf0\x9f\xa4\xa5",
     "liar": "\xf0\x9f\xa4\xa5",
     "no_mouth": "\xf0\x9f\x98\xb6",
```
Cephian, line 78: I think we should also remove `:no_mouth:` 
image

```suggestion
```
The throwing up one is pretty funny, so I say we keep that one. Other than that, looks good!

Cephian (approved):

PR #26, opened 2022-01-15
Add color to !list

Cephian:
Add color `#dd2e44` to the result of the !list command. This is the same color Discord uses for the heart part of the flaming heart emoji used in the same command.

image


anoadragon453 (approved):
Looks good to me. Also TIL GitHub automatically shows the color in the rendered comment when you paste a hex color code - neat!

PR #27, opened 2022-01-15
Format filenames more nicely

Cephian:
Currently we print song titles as the names of the files attached. To make them look nicer, we can

- Drop the file extension
- Replace underscores with spaces

I've attached an image where I did some experimentation and concluded that Discord does not allow filenames starting with underscores, so we should not get any "empty" song names this way. Currently multiple spaces seem to be consolidated to one space. I'm fairly sure this is an issue with the way Discord renders the embed and not anything in this code, but there might be a way to fix it.

I think needing to maintain the exact number of underscores as they convert to spaces is such an obscure use case we can ignore it for now (or forever).

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -97,6 +97,17 @@ async def on_message(message: Message):
         await command_stop()
 
 
+# Take a filename as string and return it formatted nicely
+def format_filename(filename):
+    # Remove file extension
+    filename = path.splitext(filename)[0]
+
+    # Replace all underscores with spaces
+    filename = filename.replace("_", " ")
+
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -97,6 +97,17 @@ async def on_message(message: Message):
         await command_stop()
 
 
+# Take a filename as string and return it formatted nicely
+def format_filename(filename):
```
anoadragon453, line 101:
```suggestion
def format_filename(filename: str):
```

types

anoadragon453 (approved):
One tiny nitpick, otherwise lgtm!

Issue #16, opened 2022-01-09
Detecting whether an attachment is media through `attachment.content_type` can sometimes fail


anoadragon453:
We have the following code for detecting whether an attachment in a Discord message is a song/video:

https://github.com/anoadragon453/busty/blob/15c4bb0d20a2160f75dbb390b8c10757c68bae67/main.py#L214-L217

However, on accidentally running `!list` on a fairly old channel with a lot of messages/attachment, I got the following stack trace:

<details>

<summary>Exception</summary>

```
 % python ./main.py 
We have logged in as Busty Dev#3334.
Ignoring exception in on_message
Traceback (most recent call last):
  File "/home/user/code/busty/env/lib/python3.10/site-packages/discord/client.py", line 343, in _run_event
    await coro(*args, **kwargs)
  File "/home/user/code/busty/./main.py", line 57, in on_message
    await list(message)
  File "/home/user/code/busty/./main.py", line 197, in list
    channel_media_attachments = await scrape_channel_media(message.channel)
  File "/home/user/code/busty/./main.py", line 232, in scrape_channel_media
    not attachment.content_type.startswith("audio")
AttributeError: 'NoneType' object has no attribute 'startswith'
```
</details>

It turns out that `Attachment.content_type` can be `None`, according to [discord.py's documentation](https://discordpy.readthedocs.io/en/latest/api.html?highlight=attachment#discord.Attachment). Since it's new in discord.py v1.7, my hunch is that Discord only started recording and returning the media type of attachments recently, and thus this field is `None` when querying older message attachments.

This doesn't really affect Busty's at all, as we're always dealing with newly uploaded media files, but it'd be good hygiene to handle, especially if my hunch is wrong.

Cephian:
I just found out by accident that the attachment type is also "None" if you send a file with no extension and this crash occurs. This seems much more likely to actually happen by accident, so we should definitely patch this.

anoadragon453:
@Cephian nice catch! Definitely worth fixing then.

anoadragon453:
Fixed by https://github.com/anoadragon453/busty/pull/28.

PR #28, opened 2022-01-15
handle if attachment.content_type is None

Cephian:
Fixes Issue #16 by handling the case when attachment.content_type is None

### DIFF
--- a/main.py
+++ b/main.py
@@ -311,9 +311,10 @@ async def scrape_channel_media(
             continue
 
         for attachment in message.attachments:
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
This lgtm. If someone uploads a song without an extension then their song won't be counted - however I'm not sure of any other way to discern that from any other non-song file. Plus I think it's unlikely that we'll get genuine song submissions without a file extension.

And if that happens, we can just poke them to re-upload.

PR #29, opened 2022-01-15
Make "Thanks for coming" message an embed

Cephian:
Uses the same embed color as the `!list` command

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -183,9 +183,12 @@ def play_next_song(e=None):
                 await bot_member.edit(nick=original_bot_nickname)
 
             # Say our goodbyes
-            await current_channel.send(
-                "Thas it y'all. Hope ya had a good **BUST** ❤️‍🔥 "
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -183,9 +183,12 @@ async def inner_f():
                 await bot_member.edit(nick=original_bot_nickname)
 
             # Say our goodbyes
-            await current_channel.send(
-                "Thas it y'all. Hope ya had a good **BUST** ❤️‍🔥 "
+            embed_title = "❤️‍🔥 Thas it y'all ❤️‍🔥"
+            embed_content = "Hope ya had a good **BUST**!"
```
anoadragon453, line 187: Should we make the `!` bold as well?

anoadragon453 (approved):

main.py
Cephian: The difference is slight but I think you're right, the bold `!` looks better
image

Issue #25, opened 2022-01-15
Read media info from tags


Cephian:
Most audio formats allow for storing tag info like Title and Artist. It would be nice if Busty could be made aware of this info to display in the `!list` and "Now Playing" messages, instead of just taking it from filenames.

mp3 tags are really all we need, but tag support for flac, wav, ogg, etc is a good idea also if it's not too much extra work.

Cephian:
Resolved by #30

PR #30, opened 2022-01-16
Solved: Read media info from tags 

DipayanP007:
# Modified Code so that:
* When the command is `!list`, the output will have audio title, artist, album in the format `title | artist | album`
* If the media does not have the specified tags, the filename will be used instead

## Screenshots:
* `!list command`
 
list

* `!play command`

 play

Solves the issue #25

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,6 +3,7 @@ import os
 import random
 from os import path
 from typing import List, Optional, Tuple
+from tinytag import TinyTag
 
 import discord
 from discord import (
@@ -99,13 +100,27 @@ async def on_message(message: Message):
 
 # Take a filename as string and return it formatted nicely
 def format_filename(filename: str):
-    # Remove file extension
-    filename = path.splitext(filename)[0]
+    
+    # Get all the tags for a track
+    audio = TinyTag.get(f"{attachment_directory_filepath}/{filename}")
 
-    # Replace all underscores with spaces
-    filename = filename.replace("_", " ")
+    # If the tag doesnot exist display the file name only
+    if len(audio.title.strip()) == 0:
+        filename = path.splitext(filename)[0]
 
-    return discord.utils.escape_markdown(filename)
+        filename = filename.replace("_", " ")
+        content = filename.replace("_", " ")
+
+    # Otherwise, display the audio title, artist, album in format "title | artist | album"
+    else:
+        content = f" {str(audio.title)}"
+    if len(audio.artist.strip()) != 0:
+        content = content + f" | {str(audio.artist)}"
+    if len(audio.album.strip()) !=0:
+        content = content + f" | {str(audio.album)}"
+
+    return discord.utils.escape_markdown(content)
+
 
 
 async def command_stop():
--- a/main.py
+++ b/main.py
@@ -3,7 +3,6 @@ import os
 import random
 from os import path
 from typing import List, Optional, Tuple
-from tinytag import TinyTag
 
 import discord
 from discord import (
@@ -17,6 +16,7 @@ from discord import (
     TextChannel,
     VoiceClient,
 )
+from tinytag import TinyTag
 
 # SETTINGS
 # How many seconds to wait in-between songs
@@ -100,7 +100,7 @@ async def on_message(message: Message):
 
 # Take a filename as string and return it formatted nicely
 def format_filename(filename: str):
-    
+
     # Get all the tags for a track
     audio = TinyTag.get(f"{attachment_directory_filepath}/{filename}")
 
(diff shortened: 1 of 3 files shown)


anoadragon453:
Thanks @DipayanP007! If you find yourself battling with the CI, note that there is a script (`scripts-dev/lint.sh`) that will fix/point out style errors locally. Thus, you can fix them before pushing a commit to GitHub.

The CI is just running the same commands as that script. So if the script passes, then the CI should to!

DipayanP007:
Hi @anoadragon453 thanks for the info. I did the same and the script ran on local. I guess the ci will pass now.

DipayanP007:
And review the changes if you can :smile:

Cephian (changes requested):
Hey Dipayan, awesome work! Thanks for the PR.

I have a few formatting suggestions/requests:

1. The "Album" field unnecessarily clutters the output, So I think it's best to just not display that one
2. Instead of the format `@user - <Artist Tag> | <Title Tag>` I think `@user: <Artist Tag> - <Title Tag>` is closer to what's more commonly used for titles, so we should prefer it. Files without tags should display `@user: <Filename>` for consistency, where the filename is processed as it is now.
3. If a user submits a song with a Title tag but no Artist tag, we should just display `@user: <Title Tag>`

With these changes I think this should be ready to commit!

anoadragon453 (commented):
Some things I noticed. Thanks again for picking this up!

DipayanP007:
Hey @anoadragon453 @Cephian I have made the changes you requested. Review it
Here is a screenshot of the current embed:
ss

main.py
```diff
@@ -99,13 +100,29 @@ async def on_message(message: Message):
...
-    # Replace all underscores with spaces
-    filename = filename.replace("_", " ")
+    # Get all the tags for a track
+    audio = TinyTag.get(
+        str(os.path.join(f"{attachment_directory_filepath}", f"{filename}"))
+    )
+    content = ""
+    # If the tag doesnot exist or is whitespace display the file name only
+    # Otherwise display in the format @user: <Artist-tag> - <Title-tag>
+    if audio.artist is not None and len(audio.artist.strip()) != 0:
```
Cephian, line 111: I missed this in my first review, but using `string.isspace()` seems better than checking if `len(string.split()) == 0`
```suggestion
    if audio.artist is not None and not audio.artist.isspace():
```

main.py
```diff
@@ -99,13 +100,29 @@ async def on_message(message: Message):
...
+    if audio.title is not None:
+        if len(audio.title.strip()) == 0:
+            filename = path.splitext(filename)[0]
+            content = content + filename.replace("_", " ")
+        else:
+            content = content + f"{str(audio.title)}"
+    # If the title tag doesnot exist but the artist tag exists, display the file name along with artist tag
+    else:
+        filename = path.splitext(filename)[0]
+        content = content + filename.replace("_", " ")
```
Cephian, line 123: We can remove some duplicated code if we write the if statements like this
```suggestion
    if audio.title is not None and not audio.title.isspace():
        content = content + f"{str(audio.title)}"
    # If the title tag does not exist but the artist tag exists, display the file name along with artist tag
    else:
        filename = path.splitext(filename)[0]
        content = content + filename.replace("_", " ")
```

Cephian (changes requested):
The code looks and runs good! I made three suggestions, each just for code style. After that, I think we should be ready to merge. Thanks!

DipayanP007:
Hey @Cephian I made the changes requested. :smile:

main.py
```diff
@@ -99,13 +100,25 @@ async def on_message(message: Message):
...
-    # Replace all underscores with spaces
-    filename = filename.replace("_", " ")
+    # Get all the tags for a track
+    audio = TinyTag.get(
+        str(os.path.join(f"{attachment_directory_filepath}", f"{filename}"))
+    )
+    content = ""
+    # If the tag does not exist or is whitespace display the file name only
+    # Otherwise display in the format @user: <Artist-tag> - <Title-tag>
+    if audio.artist is not None and not audio.artist.isspace():
```
Cephian, line 111:
```suggestion
    if audio.artist is not None and audio.artist.strip():
```

main.py
```diff
@@ -99,13 +100,25 @@ async def on_message(message: Message):
...
+        content = content + f"{str(audio.artist)} - "
+
+    if audio.title is not None and not audio.title.isspace():
```
Cephian, line 114:
```suggestion
    if audio.title is not None and audio.title.strip():
```

Cephian (changes requested):
Oops! This time it was my mistake, I recommended code without testing it. It turns out `text.isspace()` returns `False` on empty strings, so what you had was better! I fixed the two isspace instances as recommendations (and actually tested the code this time!) so with this, the PR should finally be good.

DipayanP007:
Hey @Cephian did the changes you requested

Cephian (approved):
Everything looks good! Thanks so much for your PR!

Issue #32, opened 2022-01-16
Use media info if available when displaying the currently playing song as Busty's displayname


anoadragon453:
#25 added the ability to read media information from files. #3 made it so that Busty changes her name to that of a song's artist and name before it plays.

We should update the Busty name change feature so that it pulls the media information if available.

Since we have limited characters to work with in a Discord displayname (up to 32), I suggest that we use the form:

```
{random_emoji}{file_artist_name or discord_displayname} - {file_song_name or cleaned_up_file_name}
```

Issue #31, opened 2022-01-16
Busty cannot handle two files with the same name


Cephian:
If two people submit songs with the same filename, the latter will overwrite the former in the media directory. We should save files with local filenames that don't conflict, perhaps something like `my_song.mp3` --> `<discord message id>.mp3`.

PR #33, opened 2022-01-16
Make nickname string tag-aware

Cephian:
Implements #32.
image

Works internally by generalizing previous tag-aware formatting function `format_filename` (now nameed `song_format`) to allow a "fallback artist name" which if not passed is unused. Markdown escaping is also removed from inside the function Then:

- To format correctly in !list and the next song message, we don't pass a fallback artist and markdown escape the result.
- To format correctly in the nickname we pass `author.display_name` of the song submitter as the fallback artist name.

In reorganizing the function, incidentally a parameter `local_filepath` has been added (separate from `filename`, the filename on Discord itself). The previous version of the function assumed that the local filepath was always `<media directory>/<discord filename>`. This is true for now, but will not be after Issue #31 is fixed.


main.py
```diff
@@ -98,27 +104,33 @@ async def on_message(message: Message):
...
 
-# Take a filename as string and return it formatted nicely
-def format_filename(filename: str):
+# Format a song as text nicely using artist/title tags if available
+# local_filepath is the actual path on disc
+# filename is the filename on Discord
+# artist_fallback is the fallback author value (no fallback if not passed)
+def song_format(local_filepath: str, filename: str, artist_fallback: str = ""):
+    # tag from tag list is valid as an entry
+    def valid_tag(tag):
```
anoadragon453, line 113: another candidate for a docstring conversion. though since this function is inline, we probably don't need to go all out with args/return sections.

```suggestion
    def valid_tag(tag):
        """
        Returns True if a given tag list entry is valid, otherwise False.
        
        A "valid" tag is defined as a str that contains at least one non-whitespace character.
        """
```

main.py
```diff
@@ -98,27 +104,33 @@ async def on_message(message: Message):
         await command_stop()
 
 
-# Take a filename as string and return it formatted nicely
-def format_filename(filename: str):
+# Format a song as text nicely using artist/title tags if available
+# local_filepath is the actual path on disc
+# filename is the filename on Discord
+# artist_fallback is the fallback author value (no fallback if not passed)
+def song_format(local_filepath: str, filename: str, artist_fallback: str = ""):
```
anoadragon453, line 111: I would probably default `artist_fallback` to `None` rather than an empty string. Especially as we check for `None` below.

Of course `None` isn't of the `str` type. In this case though, you can make use of the [Optional](https://docs.python.org/3/library/typing.html#typing.Optional) type. Which can be either `<some type>` or `None`. For instance: `artist_fallback: Optional[str] = None`.

main.py
Cephian: Because it's just a nested helper function, and because `song_format()`itself is so short, i think a docstring here might be overkill. I think changing it to `is_valid_tag` and just a single comment above it defining a valid tag is enough, though let me know if you disagree

PR #38, opened 2022-01-18
Allow !list with target channel

### LINKED ISSUE #37: Allow `!list <channel mention>` to allow listing of media in a channel separate from the current one
So that admins can test a `!list` before doing it publicly.

Cephian:
Closes #37.

Allows calling `!list <channel mention>`. This what the output message would be if you called `!list` in the mentioned channel, but sends the message in the same channel that the command was called from. The output of `!list` is now only pinned if the channel the command was called from is the same as the channel being listed.

### DIFF
--- a/main.py
+++ b/main.py
@@ -284,6 +284,13 @@ async def command_list(message: Message):
     # if any channels were mentioned in the message, use the first from the list
     if message.channel_mentions:
         target_channel = message.channel_mentions[0]
+    if message.channel_mentions:
+        mentioned_channel = message.channel_mentions[0]
+        if isinstance(mentioned_channel, TextChannel):
+            target_channel = mentioned_channel
+        else:
+            await message.channel.send("That ain't a text channel.")
+            return
 
     # Scrape all tracks in the target channel and list them
     channel_media_attachments = await scrape_channel_media(target_channel)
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -280,8 +280,13 @@ async def inner_f():
 
 
 async def command_list(message: Message):
-    # Scrape all tracks in the message's channel and list them
-    channel_media_attachments = await scrape_channel_media(message.channel)
+    target_channel = message.channel
+    # if any channels were mentioned in the message, use the first from the list
+    if message.channel_mentions:
+        target_channel = message.channel_mentions[0]
```
anoadragon453, line 286: Technically a user could mention a voice/stage/category channel as well. While I don't think we'll ever do this ourselves, we may as well handle the case.

```suggestion
    if message.channel_mentions:
        mentioned_channel = message.channel_mentions[0]
        if isinstance(mentioned_channel, TextChannel):
            target_channel = mentioned_channel
        else:
            await message.channel.send("That ain't a text channel.")
            return
```

main.py
```diff
@@ -300,25 +305,27 @@ async def command_list(message: Message):
...
 
-    try:
-        await list_message.pin()
-    except Forbidden:
-        print(
-            'Insufficient permission to pin tracklist. Please give me the "manage_messages" permission and try again'
-        )
-    except (HTTPException, NotFound) as e:
-        print("Pinning tracklist failed: ", e)
+    # Only pin message if command is called from the target channel
```
anoadragon453, line 312:
```suggestion
    # Only pin the message if the command is called from the target channel
```

anoadragon453 (approved):
This will be super handy pre-event. Thanks for tackling it!

LGTM otherwise.

Issue #36, opened 2022-01-18
Allow `!bust <song #>` to start a bust at a given song offset


anoadragon453:
Just in case something goes wrong halfway through, and we need to restart without typing `!skip` 15 times.

PR #42, opened 2022-01-21
Allow !bust to take a position to skip to

Subserial:
Implements #36.
Extra parsing logic is added to the !bust command. If a valid song index is provided to !bust, the command passes the number of songs to skip to the play_next_song function. That function will then remove that many songs before popping the next song to play.


main.py
```diff
@@ -79,7 +79,30 @@ async def on_message(message: Message):
             await message.channel.send("You need to use !list first, sugar.")
             return
 
-        await command_play(message)
+        # split on spaces
+        # TODO: Possibly parse arguments before casing on command
+        command_args = message.content.split(" ")
```
anoadragon453, line 84: Be careful of passing a separator value to [str.split](https://docs.python.org/3/library/stdtypes.html#str.split). As the documentation states, it changes up the splitting algorithm.

Say `message.content` is `!bust<space><space>15`. If we pass `<space>` as the `sep` argument, `str.split` will actually return:

```
>>> command = "!list  15"
>>> command.split(" ")
['!list', '', '15']
```

However, if we don't set `sep`, then `str.split` will regard consecutive instances of the separator string as a single instance. Thus we end up with:

```
>>> command = "!list  15"
>>> command.split()
['!list', '15']
```

As it's likely for people to accidentally double-space, we should probably forgo setting `sep`.

```suggestion
        command_args = message.content.split()
```

anoadragon453, line 84: To help make reasoning about this list easier, I may remove the command itself from the list, so `command_args` consists of only the index.

```suggestion
        command_args = message.content.split()[1:]
```

Though you may want to split that step out into a separate line for better readability.

main.py
```diff
@@ -137,7 +160,7 @@ async def command_stop():
         await bot_member.edit(nick=original_bot_nickname)
 
 
-async def command_play(message: Message):
+async def command_play(message: Message, skip_count=0):
```
anoadragon453, line 163: Please add types. They help IDEs give better intellisense, thus preventing bugs!

```suggestion
async def command_play(message: Message, skip_count: int = 0):
```

main.py
```diff
@@ -190,7 +213,7 @@ def command_skip():
     active_voice_client.stop()
 
 
-def play_next_song(e=None):
+def play_next_song(e=None, skip_count=0):
```
anoadragon453, line 216:
```suggestion
def play_next_song(e: BaseException = None, skip_count: int = 0):
```

Hah, I also forgot to add types. Might as well fix while we're here.

main.py
```diff
@@ -79,7 +79,30 @@ async def on_message(message: Message):
             await message.channel.send("You need to use !list first, sugar.")
             return
 
-        await command_play(message)
+        # split on spaces
+        # TODO: Possibly parse arguments before casing on command
```
anoadragon453, line 83: What do you mean by this TODO?

anoadragon453 (changes requested):
Hello @Subserial!! Thanks for tackling this one :)

This works well in my local testing, besides one bug I noticed. Some small style suggestions otherwise. Looks pretty good overall though.

main.py
```diff
@@ -79,7 +79,30 @@ async def on_message(message: Message):
             await message.channel.send("You need to use !list first, sugar.")
             return
 
-        await command_play(message)
```
anoadragon453, line 82: Oh, and could you update the README to include the new syntax?

main.py
Subserial: Ideally arguments would be split and parsed before this specific condition, but I guess there's no extra considerations necessary at the moment.

main.py
Subserial: Done :)

main.py
```diff
@@ -88,7 +88,28 @@ async def on_message(message: Message):
...
+        command_args = message.content.split()[1:]
+        skip_count = 0
+        if len(command_args) > 1:
+            bust_index = 0
```
anoadragon453, line 94:
```suggestion
```

As `bust_index` is either set below or the function returns, this line isn't necessary.

main.py
```diff
@@ -88,7 +88,28 @@ async def on_message(message: Message):
...
+            try:
+                # Expects a positive integer
+                bust_index = int(command_args[1])
```
anoadragon453, line 97:
```suggestion
                bust_index = int(command_args[0])
```

Now that `command_args` consists of only arguments.

anoadragon453 (changes requested):
Just some small, last changes - otherwise this looks good to me!

PR #44, opened 2022-01-22
Add missing documentation for new features to the README

anoadragon453:
#10 and #38 added new functionality, but we forgot to document it...

This PR adds documentation to the README for both features.

Blocked on https://github.com/anoadragon453/busty/pull/42, as both PRs are likely to conflict.


README.md
```diff
@@ -43,10 +49,10 @@ It should connect to Discord and display the currently logged-in application nam
...
 ## Command reference
 
-1. `!list` - download and list all media files in the text channel this is used in. This needs to be run before `!bust` can be.
-2. `!bust` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work.
-3. `!skip` - skips the current track :scream: 
-4. `!stop` - stop busting early :scream: :scream: :scream: 
+1. `!list [<channel>]` - download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`.
+1. `!bust` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work.
+1. `!skip` - skips the current track :scream: 
+1. `!stop` - stop busting early :scream: :scream: :scream:
```
Cephian, line 55: This was an existing issue not related to documenting the new features, but since we're editing these lines anyways we should make capitalization consistent here
```suggestion
1. `!list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`.
1. `!bust` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work.
1. `!skip` - Skips the current track :scream: 
1. `!stop` - Stop busting early :scream: :scream: :scream: 
```

Cephian (changes requested):

README.md
```diff
@@ -30,6 +30,12 @@ The complete list of environment variable configuration options is
 2. `BUSTY_COOLDOWN_SECS` - Number of seconds between songs (default = 10)
 3. `BUSTY_ATTACHMENT_DIR` - Directory to save attachments (default = attachments)
 4. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
+5. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The path to a file containing a custom list of emoji to display.
```
Cephian, line 33: I accidentally didn't save these comments in my first review, so I'm writing it again. First off, here we should show a default value (default = emoji_list) to be consistent. Second off, "the path to a file" is not technically accurate, note the default value does not have a file extension. A more accurate description would be something like "The Python module to import containing the emoji list"

```suggestion
5. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The Python module to import containing the emoji list (default = emoji_list)
```
If this change is made then line 38 should also be reworded from `the path of the new file` to something like `the name of the new module`.

README.md
anoadragon453: Looks like this is additionally solved in #42. I'll block this PR until that one merges (and fix any conflicts here).

Cephian (approved):

PR #46, opened 2022-01-22
Make nickname string tag-aware

Cephian:
Implements #32.
image

Works internally by generalizing previous tag-aware formatting function `format_filename` (now nameed `song_format`) to allow a "fallback artist name" which if not passed is unused. Markdown escaping is also removed from inside the function Then:

- To format correctly in !list and the next song message, we don't pass a fallback artist and markdown escape the result.
- To format correctly in the nickname we pass `author.display_name` of the song submitter as the fallback artist name.

In reorganizing the function, incidentally a parameter `local_filepath` has been added (separate from `filename`, the filename on Discord itself). The previous version of the function assumed that the local filepath was always `<media directory>/<discord filename>`. This is true for now, but will not be after Issue #31 is fixed.

### DIFF
--- a/main.py
+++ b/main.py
@@ -107,27 +107,51 @@ async def on_message(message: Message):
         await command_stop()
 
 
-# Take a filename as string and return it formatted nicely
-def format_filename(filename: str):
+# Format a song as text nicely using artist/title tags if available
+# local_filepath is the actual path on disc
+# filename is the filename on Discord
+# artist_fallback is the fallback artist value (no fallback if not passed)
+def song_format(
+    local_filepath: str, filename: str, artist_fallback: Optional[str] = None
+) -> str:
+    """
+    Format a song as text nicely using artist/title tags if available
 
-    # Get all the tags for a track
-    audio = TinyTag.get(
-        str(os.path.join(f"{attachment_directory_filepath}", f"{filename}"))
-    )
+    Aims for the format "Artist - Title", however if the Artist tag is not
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -107,27 +107,51 @@ async def on_message(message: Message):
...
+
+    Args:
+        local_filepath: the actual path on disc
+        filename: the filename on Discord
+        artist_fallback: the fallback author value (no fallback if not passed)
+
+    Returns:
+        A string presenting the given song information in a human-readable way.
+    """
+    # a valid tag a string with at least one non-whitespace character
```
anoadragon453, line 133:
```suggestion
    # a valid tag is a string with at least one non-whitespace character
```

main.py
```diff
@@ -107,27 +107,51 @@ async def on_message(message: Message):
         await command_stop()
 
 
-# Take a filename as string and return it formatted nicely
-def format_filename(filename: str):
+# Format a song as text nicely using artist/title tags if available
+# local_filepath is the actual path on disc
+# filename is the filename on Discord
+# artist_fallback is the fallback artist value (no fallback if not passed)
```
anoadragon453, line 113: We can probably remove this now that the docstring contains the same content.

anoadragon453 (approved):
lgtm, other than a couple tiny things.

main.py
Cephian: Ah yeah, I left those in there by accident

Issue #40, opened 2022-01-19
Upload a zip archive of all songs for a bust


anoadragon453:
So that people who want to archive the songs can do so easily.

We wouldn't be able to host the songs on Discord, but if we end up integrating with google services as part of #12, then uploading to google drive may be an option.

Perhaps either attaching it as a link to the `!list` command output, or sending it as a message at the end of a bust.

A request from Toasterless.

Cephian:
I'm against attaching it to the end of `!list` as the compress+upload time would greatly increase latency of the command.

I might be okay with sending it in the "thanks for coming" message. I don't think this is a feature I'd implement myself as it doesn't to save _that_ much time over just clicking every single direct link in the pinned message in sequence and downloading that way.

anoadragon453:
Perhaps an `!archive` command that would need to be run after `!list`.

We could just post these at the end of an archive bust channel.

PR #50, opened 2022-01-25
Save files with unique filenames

### LINKED ISSUE #31: Busty cannot handle two files with the same name

Cephian:
Closes #31.

Instead of saving files with the filename direct from Discord, use the name `(Track #).(Discord Filename)` where the track number is formatted to be exactly 3 digits. Recall that busty only reads the first 500 messages, so 3 digits is sufficient. For ex the 7th song `my_song.mp3` would be saved as `007.my_song.mp3`.

### DIFF
--- a/main.py
+++ b/main.py
@@ -446,7 +446,10 @@ async def scrape_channel_media(
 
             # Save attachment content
             attachment_filepath = path.join(
-                attachment_directory_filepath, attachment.filename
+                attachment_directory_filepath,
+                "{:03d}.{}".format(
+                    len(channel_media_attachments) + 1, attachment.filename
+                ),
             )
             await attachment.save(attachment_filepath)
 


anoadragon453:
How come you didn't go with the message ID in the end? That seemed suitably unique?

Cephian:
message IDs are indeed suitably unique, but I thought this might be better since
1) A human can tell at a glance which file is which, for potential debugging situations
2) If #40 is implemented we'd need a human-readable way to deal with duplicate filenames anyways, and this is nice because it also includes bust order (something lost with just a collection of files)
3) No information is lost about the original filename is lost (again useful for both archival and debugging)

anoadragon453 (approved):
I agree with each of those reasons for the new formatting; it's a much nicer design. This looks good to me!

PR #51, opened 2022-01-25
Read cover art from audio files

### LINKED ISSUE #48: Show file cover art in "Now playing" embed
Many audio formats allow for embedded cover art, and Discord embeds have a special image field. It would be great if Busty could parse this cover art tag and display it during the "Now Playing" embed

Helpeful Links:
[Parsing image tags with TinyTag](https://pypi.org/project/tinytag/).
[Using local image files for Discord embeds](https://discordpy.readthedocs.io/en/latest/faq.html#how-do-i-use-a-local-image-file-for-an-embed-image).

Cephian:
Closes #48.
image

Rather than allowing users to send as large a file as they'd like, cover art filesize is capped to 8MB, or 8 million bytes. This is less than the actual Discord cap of (about) 8MiB, though the Discord cap CAN rise if the bot is in a boosted server.

Since the time take to upload the cover image does cause message latency, and since there's really no reason anyone needs cover art over 1MB let alone 8, I think this is a good cap (I'd even be ok lowering it).

This commit adds a dependency to Pillow, solely because we need to know which file format the cover art is, and it seems neither Discord nor TinyTag can detect this themselves. We are using such a small part of the Pillow library that it shouldn't be too difficult to eliminate this dependency by checking some magic bytes ourselves, but I propose that we accept the dependency for now and remove it later if we feel like it.

### DIFF
--- a/main.py
+++ b/main.py
@@ -18,6 +18,7 @@ from discord import (
     VoiceClient,
 )
 from PIL import Image
+from PIL import UnidentifiedImageError
 from tinytag import TinyTag
 
 # CONSTANTS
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -154,6 +156,20 @@ def is_valid_tag(tag: Optional[str]) -> bool:
...
+    # get image data as bytes, make sure it doesn't go over 8MB
+    # This is a safe lower bound on the Discord upload limit of 8MiB
+    image_data = tags.get_image()
+    if image_data is None or len(image_data) > 8_000_000:
+        return None
+    image = Image.open(BytesIO(image_data))
+    cover_filename = f"cover.{image.format}".lower()
+    cover_filepath = path.join(attachment_directory_filepath, cover_filename)
+    image.save(cover_filepath)
+    return discord.File(cover_filepath, filename=cover_filename)
```
anoadragon453, line 170: We can actually upload the bytes as a file without needing to save it to disk first, by reusing our `BytesIO` object:

```python
def get_cover_art(filename: str) -> Optional[discord.File]:
    # Get image data as bytes
    tags = TinyTag.get(filename, image=True)
    image_data = tags.get_image()

    # Make sure it doesn't go over 8MB.
    # This is a safe lower bound on the Discord upload limit of 8MiB
    if image_data is None or len(image_data) > 8_000_000:
        return None

    # Get a file pointer to the bytes
    image_bytes_fp = BytesIO(image_data)

    # Read the filetype of the bytes and discern the appropriate file extension
    image = Image.open(image_bytes_fp)

    # Wind back the file pointer in order to read it a second time
    image_bytes_fp.seek(0)

    # Make up a filename
    cover_filename = f"cover.{image.format}".lower()

    # Create a new discord file from the file pointer and name
    return discord.File(image_bytes_fp, filename=cover_filename)
```

main.py
```diff
@@ -154,6 +156,20 @@ def is_valid_tag(tag: Optional[str]) -> bool:
...
 
 
+def get_cover_art(filename: str) -> Optional[discord.File]:
+    tags = TinyTag.get(filename, image=True)
+    # get image data as bytes, make sure it doesn't go over 8MB
+    # This is a safe lower bound on the Discord upload limit of 8MiB
+    image_data = tags.get_image()
+    if image_data is None or len(image_data) > 8_000_000:
+        return None
+    image = Image.open(BytesIO(image_data))
```
anoadragon453, line 166: I tried, and it's possible to embed a non-image file inside of an mp3. Attempting to load this into Pillow will raise an `PIL.PIL.UnidentifiedImageError` error. We should catch this (ideally log it) and just return `None` if that happens.

anoadragon453 (changes requested):

anoadragon453 (approved):

PR #56, opened 2022-01-25
Update requirements (discord.py --> nextcord)

### LINKED ISSUE #53: Migrate from discord.py
Since discord.py is [discontinued](https://github.com/Rapptz/discord.py), we should probably switch to a maintained fork. The two to choose from seem to be [nextcord](https://github.com/nextcord/nextcord) and [pycord](https://github.com/Pycord-Development/pycord). I'm in favor of pycord since It seems to have a lot more attention and active development from the Github.

pycord is in theory a drop-in replacement (we'd only need to change requirements.txt) and nextcord would be _almost_ drop in (the import changes to "nextcord"). @anoadragon453 If you are okay with switching to pycord I can make sure all our old code runs fine with it and submit the pull request.

Cephian:
Closes #53. While nextcord recommends changing the import from "import discord" to "import nextcord" it's not required, and I have not done so here. As far as I can tell they don't plan on removing this alias anytime soon, so I think think should keep it for now in case we need to move back to discord.py or over to pycord.

Incidentally, I've also changed the TinyTag dependency from =1.7.0 to >=1.7.0 to be in line with the other dependencies which all have only a lower bound.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ And install the dependencies:
 pip install -r requirements.txt
 ```
 
+To install additional packages for speedup, install the `nextcord[speed]` package as well.
+
 ## Configure
 
 You'll need to create a Discord app, add a bot component, and copy the bot token.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 nextcord[voice]>=2.0.0a6
-nextcord[speed]>=2.0.0a6
 tinytag>=1.7.0
 pillow>=9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pynacl>=1.4.0
-discord.py>=1.7.3
-tinytag==1.7.0
+nextcord>=2.0.0a6
+tinytag>=1.7.0
 pillow>=9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pynacl>=1.4.0
-nextcord>=2.0.0a6
+nextcord[voice]>=2.0.0a6
+nextcord[speed]>=2.0.0a6
 tinytag>=1.7.0
 pillow>=9.0
--- a/main.py
+++ b/main.py
@@ -20,8 +20,8 @@ from nextcord import (
     NotFound,
     TextChannel,
     VoiceClient,
-    utils,
 )
+from nextcord.utils import escape_markdown
 from PIL import Image, UnidentifiedImageError
 from tinytag import TinyTag, TinyTagException
 
@@ -159,7 +159,6 @@ def song_format(
     Returns:
         A string presenting the given song information in a human-readable way.
     """
-    print('formatting the song...')
     # a valid tag is string with at least one non-whitespace character
     def is_valid_tag(tag: Optional[str]) -> bool:
         return tag is not None and tag.strip()
@@ -350,7 +349,7 @@ def play_next_song(e: BaseException = None, skip_count: int = 0):
         list_format = "{0}: [{1}]({2}) [`↲jump`]({3})"
         embed_content = list_format.format(
             submit_message.author.mention,
-            utils.escape_markdown(
+            escape_markdown(
                 song_format(local_filepath, attachment.filename)
             ),
             attachment.url,
--- a/main.py
+++ b/main.py
@@ -349,9 +349,7 @@ def play_next_song(e: BaseException = None, skip_count: int = 0):
         list_format = "{0}: [{1}]({2}) [`↲jump`]({3})"
         embed_content = list_format.format(
             submit_message.author.mention,
-            escape_markdown(
-                song_format(local_filepath, attachment.filename)
-            ),
+            escape_markdown(song_format(local_filepath, attachment.filename)),
             attachment.url,
             submit_message.jump_url,
         )
@@ -374,9 +372,7 @@ def play_next_song(e: BaseException = None, skip_count: int = 0):
             await current_channel.send(embed=embed)
 
         # Play song
-        active_voice_client.play(
-            FFmpegPCMAudio(local_filepath), after=play_next_song
-        )
+        active_voice_client.play(FFmpegPCMAudio(local_filepath), after=play_next_song)
 
         # Change the name of the bot to that of the currently playing song.
         # This allows people to quickly see which song is currently playing.
--- a/main.py
+++ b/main.py
@@ -5,17 +5,22 @@ from io import BytesIO
 from os import path
 from typing import List, Optional, Tuple
 
-import discord
-from discord import (
+from nextcord import (
     Attachment,
     ChannelType,
+    Client,
     ClientException,
+    Embed,
+    FFmpegPCMAudio,
+    File,
     Forbidden,
     HTTPException,
+    Intents,
     Message,
     NotFound,
     TextChannel,
     VoiceClient,
+    utils,
 )
 from PIL import Image, UnidentifiedImageError
 from tinytag import TinyTag, TinyTagException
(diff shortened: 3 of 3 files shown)


anoadragon453:
Testing this locally, I found that importing things directly from `nextcord`, such as the below:

https://github.com/anoadragon453/busty/blob/e65b46cae2157120aef195eb8b887fbcfc612c7e/main.py#L9-L19

does not work. Instead, these items must be referred to directly (`from nextcord.types.message import Attachment`). They're not re-exported under the `discord` module provided by nextcord. Doing so in e65b46cae2157120aef195eb8b887fbcfc612c7e allowed the bot to work, and it seems to function correctly.

It's not a big deal to fix the imports, though I figured we were trying to avoid that in this PR.

Cephian:
@anoadragon453  It turns out that I had discord.py installed systemwide when I was "testing" nextcord, so all my tests were just with discord.py itself. Another option (I'm slightly in favor of, though I could understand if you're not) is to remove the `discord.` prefix from everything, and reduce the "nextcord" phrase within main.py down to a single instance. Namely, the import becomes 
```python
from nextcord import (
    Attachment,
    ChannelType,
    Client,
    ClientException,
    Embed,
    FFmpegPCMAudio,
    File,
    Forbidden,
    HTTPException,
    Intents,
    Message,
    NotFound,
    TextChannel,
    VoiceClient,
    utils,
)
```
with `Client`, `Embed`, `FFmpegPCMAudio`, `File`, `Intents`, and `utils` being new. It feels a little strange take something as specific as `discord.utils` and alias it to the very general seeming `utils`, but I'm not terribly worried about it since our project is so small and so inextricably linked with Discord anyways. If you don't like this, then I'm fine with the `import nextcord as discord` solution as well.

As an interesting side note, I found that with nextcord when I kill Busty from my terminal she leaves the voice chat instantly, whereas with discord.py she'd hang there until she timed out.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 pynacl>=1.4.0
-discord.py>=1.7.3
-tinytag==1.7.0
+nextcord>=2.0.0a6
```
anoadragon453, line 2: We should install the `voice` package group and drop `pynacl` from the list. `nextcord[voice]` installs pynacl, though it may require more things in the future as well. `pynacl` was only included explicitly, as I had done so in a previous discord bot project - possibly back before `discord.py` had a `voice` package group.

As for the [`speed` group](https://github.com/nextcord/nextcord#installing), I see little downside to it. I tested it, and everything continues to work fine.

The only reason I can think that it'd be optional is that some of the libraries aren't as widely supported across all platforms. In which case, we should probably recommend people install it in our docs (via `pip install nextcord[speed]`) but not require it here.

anoadragon453 (commented):
After recreating my virtualenv, `import discord` looks to work fine for me as well. I'll just chalk it up to interactions between aliases and python's cache.

The majority of those are fine I think, though `utils` does bug me. We could just do:

```python
import nextcord
...
nextcord.utils
```

perhaps? It would also make it clearer to contributors that we're using nextcord (vs discord.py, pycord).

> As an interesting side note, I found that with nextcord when I kill Busty from my terminal she leaves the voice chat instantly, whereas with discord.py she'd hang there until she timed out.

Oh yeah! I see the same. That's super handy.

Cephian:
Looking back, we only use discord.utils in one place, for escape_markdown. how about instead just `from nextcord.utils import escape_markdown` ? this way the nextcord references are next to each other in the code, and there are no package aliases or overly specific classes/modules being given general names.

anoadragon453:
> how about instead just from `nextcord.utils import escape_markdown` ?

Sounds good to me.

I've noticed some extra commits ended up here. How are you pulling from main?

Typically I'll just do `git pull origin main` on my PR branch in order to merge in the latest changes.

Cephian:
I'm just using `git pull` from the `main` branch, then switching to `nextcord` branch and using `git rebase` and then `git push`. I don't see any "extra" commits right now, you can see above I changed the base branch and then changed it back after reading [this StackOverflow comment](https://stackoverflow.com/a/46782679), though this hardly seems like the official solution. It seems like something is happening where while I rebase on main locally, Github.com still thinks I'm based on the original version of `main` I branched off from.

anoadragon453:
I was seeing https://github.com/anoadragon453/busty/commit/e0b2bce2ad1531f1a3296c72088417197f5f5f37 and https://github.com/anoadragon453/busty/commit/465d44fe210b718f89069a17d1a26611ebb4a6af in the timeline, though they appear to be gone now.

Rebasing your commits shouldn't be necessary unless you want to rewrite your PR's commit history - which isn't recommended once a review has started, as it makes the changes since a last review harder to pick out.

With squash + merge, github will squash all commits of your branch together into one, and then apply it as a new commit on top of the base branch. As long as the PR branch can merge into the base branch, this should produce the expected result, and a clean commit history.

If you do want to merge the base branches changes into your PR branch, I'd recommend merging the base branch into your PR:

```sh
git switch main
git pull
git switch my-pr
git merge main
```

`git pull origin main` from your PR branch is the equivalent, and less commands:

```sh
git pull origin main
# fix conflicts if necessary
git commit # merge commit messages are autogenerated
git push
```

If you do really want to rebase the beginning of your branch on top of the base branch, you can add the `--rebase` flag: `git pull origin main --rebase` from your PR branch.

GitHub recognises and only displays relevant changes for merge commits from the base branch on PRs: https://github.com/anoadragon453/busty/pull/44/commits/dc53b5ab4668109d36a7257e6a4274946ee095e4

requirements.txt
```diff
@@ -1,4 +1,4 @@
-pynacl>=1.4.0
-discord.py>=1.7.3
-tinytag==1.7.0
+nextcord[voice]>=2.0.0a6
+nextcord[speed]>=2.0.0a6
```
anoadragon453, line 2: (note that this can also be written as `nextcord[voice,speed]>=2.0.0a6`)

As per https://github.com/anoadragon453/busty/pull/56#discussion_r793612421, I think we should keep `speed` separate for now, and only recommend it in the documentation if people want a speed boost.

anoadragon453 (commented):
Code looks good otherwise!

`from nextcord.utils import escape_markdown` is a nice solution.

Cephian:
I see, I was originally under the impression that I should just "rebase whenever I can" so that I could incrementally make changes to get it to merge as I needed to, but I suppose thinking about it more as long as Github says the merge can be done automatically there's really no reason to do that, especially since it changes commit hashes. That makes my original worry about something like `git pull origin main` causing "useless commits" not a real issue, since in theory whenever I merge main into a branch I should be doing it to fix a specific conflict. Thanks for the explanation!

requirements.txt
Cephian: I'm fine with keeping speed separate, though my guess is that it's not separate because the required libraries are widely unsupported but instead because if you install nextcord without [voice] or [speed] it becomes dependency-less, which makes it seem cool and clean.

Cephian: Actually, I might be wrong on this. it seems [aiodns](https://pypi.org/project/aiodns/) does not list OSX as a target platform, whereas all the other libraries do.

anoadragon453: The history behind these options seems to be that discord.py just added the extra `speed` option as a lever to switch between Python's standard json lib and `orjson`: https://github.com/Rapptz/discord.py/commit/88d825a803917ab3e0e9797df9f47be7ed79bb9c. Not much public discussion happened here - it was just pushed to master.

nextcord then added the extra libraries in https://github.com/nextcord/nextcord/pull/78. `Brotli` at least is just python bindings, and depends on a `brotli` package being installed on the system.

README.md
```diff
@@ -41,6 +41,8 @@ And install the dependencies:
 pip install -r requirements.txt
 ```
 
+To install additional packages for speedup, install the `nextcord[speed]` package as well.
```
anoadragon453, line 44:
```suggestion
To install an additional set of python dependencies for speeding up the bot, install the `nextcord[speed]` package as well. You may need to install additional system packages.
```

anoadragon453 (approved):
Oops, I think I implied an approval in my last review on accident -- probably should've hit request changes instead. Sorry about that!

Other than this one final small change though, this lgtm. Feel free to commit it directly to main.

Issue #52, opened 2022-01-25
Cleanup files in the attachments folder


Cephian:
As of now Busty does not delete any attachment files, leaving them to accumulate. In theory, if run without maintenance indefinitely she would eat up more and more drive space even though once a bust is done she won't access any files in the attachments folder without re-downloading. I think Busty should keep track of which files she has created each bust and then delete them all afterwards.

I'm against an implementation which blindly deletes ALL files in the attachments folder after each bust. We should be careful to remove only files we have created.

PR #58, opened 2022-01-27
Remove attachments from disk on end of bust

Subserial:
Implements #52 
A copy of list of attachments is generated when running `!list`. This list is held until the current bust ends, in which the list is enumerated and each attachment is removed from disk. This list is overwritten if `!list` is called multiple times before `!bust`.

### DIFF
--- a/main.py
+++ b/main.py
@@ -43,9 +43,9 @@ dj_role_name = os.environ.get("BUSTY_DJ_ROLE", "bangermeister")
 # The channel to send messages in
 current_channel: Optional[TextChannel] = None
 # The media in the current channel
-current_channel_content: Optional[List] = None
+current_channel_content: Optional[List[Tuple[Message, Attachment, str]]] = None
 # The local filepaths of media from the current bust
-current_bust_content: Optional[List] = None
+current_bust_content: Optional[List[str]] = None
 # The actively connected voice client
 active_voice_client: Optional[VoiceClient] = None
 # The nickname of the bot. We need to store it as it will be
--- a/main.py
+++ b/main.py
@@ -44,7 +44,7 @@ dj_role_name = os.environ.get("BUSTY_DJ_ROLE", "bangermeister")
 current_channel: Optional[TextChannel] = None
 # The media in the current channel
 current_channel_content: Optional[List] = None
-# The media from the last bust
+# The local filepaths of media from the current bust
 current_bust_content: Optional[List] = None
 # The actively connected voice client
 active_voice_client: Optional[VoiceClient] = None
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -483,6 +500,9 @@ async def command_list(message: Message):
     global current_channel_content
     current_channel_content = channel_media_attachments
 
+    global current_bust_content
+    current_bust_content = channel_media_attachments.copy()
```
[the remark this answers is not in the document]
Cephian: If we do this, i think describing them as "local files" is more fitting than "attachments" in general (and it's how we describe them in other places in the code), since they no longer have any link to Discord structures. So probably `remove_attachments()` should be renamed to `remove_local_files` or something as well

main.py
```diff
@@ -302,9 +305,12 @@ async def inner_f():
             )
             await current_channel.send(embed=embed)
 
+            remove_attachments(current_bust_content)
```
Cephian, line 308: Please add comment `# Remember to always clean up after you bust`

Cephian (changes requested):
I'm bad at Github so I accidentally put my requested changes above as comments instead of here

Cephian (changes requested):
EDIT: (Please ignore this, Github somehow submitted my review four times)

Cephian (changes requested):
(Please ignore this, Github somehow submitted my review four times)

Cephian (changes requested):
I'm bad at Github so I accidentally put my requested changes above as comments instead of here

Cephian:
I'm very bad at github,  I submitted the same review 4x because it kept telling me it had a server error when I tried to submit it

main.py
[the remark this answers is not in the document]
Subserial: Reducing the footprint made the delete operation trivial, so I just folded it into the single call site.

main.py
```diff
@@ -281,6 +283,7 @@ def command_skip():
 def play_next_song(e: BaseException = None, skip_count: int = 0):
     async def inner_f():
         global current_channel_content
+        global current_bust_content
```
anoadragon453, line 286: We could derive this from `current_channel_content` as well, without adding another global variable?

```python
for _, local_filepath, _ in current_channel_content:
    os.remove(local_filepath)
```

Subserial: current_channel_content is treated like a queue, where the next song is popped off the front. Its empty state signals the end of the bust. I wasn't sure if this was preferred so I left that pattern untouched.

anoadragon453: Aha! We do pop from it, I had forgotten that. Definitely not something to rely upon then!

main.py
```diff
@@ -44,6 +44,8 @@
 current_channel: Optional[TextChannel] = None
 # The media in the current channel
 current_channel_content: Optional[List] = None
```
anoadragon453, line 46: And while we're here.

```suggestion
current_channel_content: Optional[List[Tuple[Message, Attachment, str]]] = None
```

main.py
```diff
@@ -44,6 +44,8 @@
...
+# The local filepaths of media from the current bust
+current_bust_content: Optional[List] = None
```
anoadragon453, line 48: for deeper type inspection.

```suggestion
current_bust_content: Optional[List[str]] = None
```

Cephian (approved):

Issue #12, opened 2022-01-07
Autogeneration of Google Forms for Voting


Cephian:
It would be convenient if Busty could automatically generate Google Forms for song voting.

From a bit of research, it seems the standard way to do this is to use [Google Apps Script](https://developers.google.com/apps-script/reference/forms). Google Apps Script code is only run remotely on Google Servers and not locally, however scripts can be created and deployed locally through the [REST API](https://developers.google.com/apps-script/api/concepts).

I plan on adding this feature myself, hopefully starting within the next week or so.

Cephian:
A much easier partial implementation of this feature could work like this:

Someone calls `!makeform` on the desired channel, and they are DMed a message with some generated google appscript code and a link to https://script.google.com/home/projects/create. Then to create the form, the user just needs to copy and paste the code and deploy it.

This is much easier than creating the form manually, but also requires no scary things like Google Authentication. Plus, almost no effort implementing this would be wasted if we were to bump it to full automatic creation.

anoadragon453:
@Cephian nice idea! It certainly makes the implementation in the bot much simpler, as well as removes the need to provide credentials to the bot - and allows users to inspect and edit the script beforehand if desired. +1!

anoadragon453:
Some ideas to speed this up. Apparently there's a 3 operation / second rate limit.

- Use the appscript REST API to eliminate the need to create the form manually.
- Instead of using multiple operations for "create question", "set required", "set title", just initially create a question with all the required attributes.


Also we probably don't need to code golf since Discord now supports large text files embedded in messages.

anoadragon453:
https://github.com/anoadragon453/busty/pull/119 goes a ways to making this better, though still takes a few tens of seconds to generate the form. Reducing the number of operations to the form could still help speed this up.

It remains to be seen whether we still want to invest in using the API directly from the bot.

PR #76, opened 2022-02-03
Add reminder to vote

### LINKED ISSUE #72: Add a "remember to vote on [the form](link)!" to the song intermission message
Specifically this message:

image

So that we don't need to do so manually.

Cephian:
Closes #72. Adds a reminder to vote onto the "Now Chillin'" message which plays before every song.

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -387,7 +387,7 @@ def play_next_song(e: BaseException = None, skip_count: int = 0):
         # Wait some time between songs
         if seconds_between_songs:
             embed_title = "Currently Chillin'"
-            embed_content = "Waiting for {} second{}...".format(
+            embed_content = "Waiting for {} second{}...\n\n**REMEMBER TO VOTE ON THE GOOGLE FORM!**".format(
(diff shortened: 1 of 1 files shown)


Cephian:
I just realized that this does not include a link to the form as #72 initially suggested. To do that we would most likely have to pass this in while busting, or wait until #12 is complete. We could also add a `!formlink` command or something which we can run in any channel during the bust, setting an internal `form_link` variable, and have the "chillin" message check to see if this is `None` or not each time. That way we don't have to remember to pass it in on `!bust`, which I could def see myself forgetting.

Or we could ignore that part for now.

anoadragon453 (approved):
A `!formlink` command sounds like a nice idea. Busty could even print out an embed saying to "come get the form: <link>" or whatever and pin it. Otherwise we're just dropping the link into the channel ourselves, which can get a bit lost in the noise.

We can do that separately though. For now, most people will have the form link from somewhere, and a reminder is all they need.

Issue #19, opened 2022-01-15
Bold the title of the currently playing song in the pinned song listing


anoadragon453:
Busty currently changes her name to the currently playing song, so that it's easy to see at a glance.

As an additional nice-to-have, we could also bold the title of the currently playing song in the song listing message. Like so:

![image](https://user-images.githubusercontent.com/1342360/149627385-641559d8-cbf9-4a7f-b859-35442613633d.png)

DipayanP007:
Hi @anoadragon453 can i work on this?

Cephian:
> Hi @anoadragon453 can i work on this?

Hey, if you'd like to work on this then go right ahead! How did you find this project by the way?

DipayanP007:
I really found the project interesting and innovative. And for the record I am workingon #25  too which is partly solved. So assign me that and I will get it done 😄

Cephian:
> I really found the project interesting and innovative. And for the record I am workingon #25 too which is partly solved. So assign me that and I will get it done smile

Sure, if you'd like to take that on as well I don't believe anyone else is working on it

anoadragon453:
Thanks for the PR @DipayanP007! We'll get to reviewing it very soon.

In the meantime, we're really grateful for your help on this project. It's allowed us to move faster on it. Given that, would you be interested in joining the Discord server that this bot is used in? It's primarily a place where people create and share music that they make every 2 weeks in a friendly competition scenario. After submissions are due, we all jump in a voice call and listen to what people have created! 

If you're interested, let us know and we'll share an invite. If not, no worries!

DipayanP007:
Hey, I would love to be part of the server. But for your information, I am not that active on discord. If it's still okay with you, you can share the invite.

anoadragon453:
@DipayanP007 no worries! Participation is up to you: `discord . gg / tJTC Q7kB` (remove spaces)

DipayanP007:
Hey @anoadragon453 `!list` now does not work when a song is being played. So do you guys still need this feature?

Cephian:
You might be a little confused about what the `!list` command it supposed to do. It's only supposed to run once to gather and list all the songs, then pin those messages. The idea is that if someone wants to see what song is currently playing and where it is in the list, they will look at the pinned messages and see it there. So the bot should keep track of the pinned list message(s) we sent at the start, and then edit them to bold/unbold a song each time one starts/stops playing.

[...]

PR #79, opened 2022-02-05
Pin currently playing song for its duration

### LINKED ISSUE #63: Temporarily pin "Now playing" message
Temporarily pin the "now playing" message for the duration of the song (and unpin it afterwards) so that people can check channel pins and see the cover art and More Info easily at any point. Right now the Now Playing message can be buried quickly when chat is very active.

Cephian:
Closes #63.

It wasn't 100% settled whether this was a good idea or not, however I think it is. It's true that it makes the pinned result of the `!list` command slightly harder to view while another song is playing. I think this is not a huge deal, since

1. Users already have a tracklist they are following along with if they are voting as they go (which voting stats seem to suggest most people do)
2. It takes minimal effort to scroll down slightly and view the tracklist, and the majority of the time the pinned message view is "stable" (e.g. not actively changing from new pins/unpins)
3. Since tracks can have album art and more info, there is a lot to be gained by being able to view them during a bust and not just for a fleeting moment at the start
4. Busty's nick can provide some at-a-glance info, but large collab songs often have long names with critical info cut off, and it can be difficult to find which song is playing at a glance in the list (granted #19 would help with this somewhat)

I consider the pinned tracklist to be more for initial and post-bust reference than mid-bust reference. I recommend that we implement this feature, and then next bust we tell people about it and ask them if we should keep it or remove it.


anoadragon453 (changes requested):
Tested, and it works! You're right in that one really only occasionally checks pins for the current spot in the overall list of songs. So the chance of pins shifting while someone is viewing them is slim, and even in that case it's not a big deal.

Cephian:
Removing the `inner_f` was a good change style-wise but was so grand it made the diffs go wonky (and will probably mandate manual refactoring on most pending PR's), so I'll summarize the changes here manually myself.

1. Printed error if `ffmpeg_post_hook` was invoked with one
2. Added async function `try_set_pin(message, pin_state)` which pins/unpins while handling error and permissions logging. call `try_set_pin(message, True)` to pin it, and `try_set_pin(message, False)` to unpin.
3. removed `inner_f` wrapper function and called play_next_song as async from `ffmpeg_post_hook`.

PR #80, opened 2022-02-05
Run cleanup tasks on bot shutdown

### LINKED ISSUE #68: Reset nickname if bot is killed
If the busty process is sent a signal to close during a bust, she will ~~leave behind files in the attachments folder that need to be cleaned out manually, as well as~~ keep whatever nickname she had at the time. We should use the [atexit](https://docs.python.org/3/library/atexit.html) module to attempt to revert Busty's nickname and delete files she is tracking before the program ends.

EDIT: We can't use the atexit module, as the discord client event loop is closed before atexit runs. We would probably have to reset the nick through some nextcord event hook.

anoadragon453:
Closes #68.

A bot can override the `close` method of a `nextcord.Client` in order to run cleanup tasks on client shutdown.

Implementing this function will override nextcord's own `Client.close()`; the implementation of which is shown here (and explains how nextcord will disconnect from VC's on shutdown): https://github.com/nextcord/nextcord/blob/8d82a0858f8440c2b62f92b8e8e6d79b6873c684/nextcord/client.py#L637-L642

Overriding this function has the benefit of allowing us to run our own code on shutdown, but also means that the original `close` method linked about *doesn't* run. I've attempted to run this manually, though this oddly leaves to infinite recursion and strange behaviour.

Ideally we'd also run the original `close` method. But for now, taking the vc disconnect feature seems to result in the desired behaviour.

This PR also pulls the bot nickname restoration and attachment cleanup tasks into their own helper methods.

### DIFF
--- a/main.py
+++ b/main.py
@@ -88,7 +88,7 @@ async def on_ready():
 async def close():
     """Perform some cleanup tasks on bot shutdown"""
     try:
-        # Restore the original nickname if it was changed.
+        # Restore the original nickname if it was changed
         await maybe_restore_original_nick()
 
         # Delete any downloaded attachments from the last bust
@@ -111,8 +111,8 @@ async def maybe_restore_original_nick():
 
 def delete_saved_attachments():
     """Delete any attachments saved during the current bust."""
-    # This will run at any point the bot shuts down, even when outside
-    # outside of a bust when current_bust_content is None
+    # This will run at any point the bot shuts down. Note that outside of a bust,
+    # current_bust_content is None
     if not current_bust_content:
         return
 
--- a/main.py
+++ b/main.py
@@ -85,6 +85,44 @@ async def on_ready():
 
 
 @client.event
+async def close():
+    """Perform some cleanup tasks on bot shutdown"""
+    try:
+        # Restore the original nickname if it was changed.
+        await maybe_restore_original_nick()
+
+        # Delete any downloaded attachments from the last bust
+        delete_saved_attachments()
+
+        # Disconnect from voice channels
+        for voice in client.voice_clients:
+            await voice.disconnect(force=True)
+
+    except Exception as e:
+        print("Exception during shutdown:", e)
+
+
+async def maybe_restore_original_nick():
+    """Restore the bot's original nickname in the guild if it was changed."""
+    if original_bot_nickname and current_channel:
+        bot_member = current_channel.guild.get_member(client.user.id)
+        await bot_member.edit(nick=original_bot_nickname)
+
+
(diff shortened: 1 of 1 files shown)


Cephian (approved):
I notice this PR abstracts a lot of the same code as PR #78 as they both do some separation of the same cleanup tasks into their own methods. Whichever we merge first, the other will have to be reorganized. I am totally fine merging this one first and rebasing #78 on this PR, but it's up to you.

I couldn't find any issues with the code. The only question I really have is whether we need to log every single time we deleted a local attachment file, but I suppose it's better to be safe than sorry and catch errors quickly when deleting files off the user's system.

anoadragon453:
> Whichever we merge first, the other will have to be reorganized. I am totally fine merging this one first and rebasing https://github.com/anoadragon453/busty/pull/78 on this PR, but it's up to you.

Hmm, they do indeed overlap a bit. I don't quite think we could run `finish_bust` inside of `close` as I don't think we'd want to print a "❤️‍🔥 Thas it y'all ❤️‍🔥" on shutdown. Perhaps extract [these lines](https://github.com/anoadragon453/busty/pull/78/files#diff-b10564ab7d2c520cdd0243874879fb0a782862c3c902ab535faabe57d5a505e1R352-R364) to a `cleanup` function, which I can then just simply run in `close` in this PR.

> The only question I really have is whether we need to log every single time we deleted a local attachment file, but I suppose it's better to be safe than sorry and catch errors quickly when deleting files off the user's system.

I think it's useful to have a record of what the bot deleted just in case something unexpected happened. It is a bit verbose, and we could solve that by having multiple log levels, but I'm not convinced we need that with the current (still quite small) size of the project.

Cephian:
> Perhaps extract [these lines](https://github.com/anoadragon453/busty/pull/78/files#diff-b10564ab7d2c520cdd0243874879fb0a782862c3c902ab535faabe57d5a505e1R352-R364) to a `cleanup` function in your PR, which I can then just simply run in `close` in this PR.

So should I do that before we merge this PR and then you'll rebase/merge in? Or do you want to merge this first since it's already reviewed and then I can make the changes on my end?

anoadragon453:
I think it'd be easier if we got your PR in first. I'll give it a review after the cleanup functions are factored out, if that sounded reasonable?

Cephian:
@anoadragon453 Alright, I've tested pushed to #78 with a factored out cleanup() function

anoadragon453:
Closing this as it will need large changes after #112 and #78. A similar change can be made to fix #68 after #78 is merged.

PR #82, opened 2022-02-06
Fix Busty raising an exception whenever she is direct messaged

Cephian:
`message.author` is of class `Member` if the message was sent in a server, and of class `User` if the message was sent via a DM. Only the `Member` class has the field `roles`,  but in `on_message()` we access `message.author.roles` without checking to see if the type is `Member` or `User`.

### DIFF
--- a/main.py
+++ b/main.py
@@ -21,6 +21,7 @@ from nextcord import (
     Forbidden,
     HTTPException,
     Intents,
+    Member,
     Message,
     NotFound,
     TextChannel,
@@ -87,6 +88,9 @@ async def on_ready():
 @client.event
 async def on_message(message: Message):
     if message.author == client.user:
+        return
+
+    if not isinstance(message.author, Member):
         return
 
     for role in message.author.roles:


main.py
```diff
@@ -89,6 +90,9 @@ async def on_message(message: Message):
     if message.author == client.user:
         return
 
+    if not isinstance(message.author, Member):
+        return
```
anoadragon453, line 94: `message.author` being a `User` is a side effect of the message appearing outside of a guild (i.e. `message.author.roles` is needed to check the user is authorised in the guild. 

Currently we don't allow any commands over DM, which is really the reason we're returning early here.

Essentially, we might want to be more explicit and check whether `message.guild` is `None`, rather than have the check revolve around a side effect.

main.py
```diff
@@ -89,6 +90,9 @@ async def on_message(message: Message):
     if message.author == client.user:
         return
 
+    if not isinstance(message.author, Member):
```
anoadragon453, line 93: Could you add a comment to explain why we'd want to exit early in this case? It may not be clear on first glance.

anoadragon453 (changes requested):
Definitely fixes the bug! Just a couple small notes on the fix below.

PR #90, opened 2022-03-11
Google Form generation with Apps Script

Cephian:
Partially implements #12. The `!form` command (when run after `!list`) prints out Google Apps Script code which can be copied and pasted to automatically generate a voting form from the Apps Script code editor in your browser.

There is still in theory work to be done on this front, for example automatically generating the responses form and statistics, or even using the Google API to generate the form without any manual copying and pasting. However, this implements 95% of what would be nice to have, and is a self-contained feature.

### DIFF
--- a/main.py
+++ b/main.py
@@ -139,6 +139,12 @@ async def on_message(message: Message):
 
         await command_play(message, skip_count)
 
+    elif message_text.startswith("!form"):
+        if not current_channel_content or not current_channel:
+            await message.channel.send("You need to use !list first, sugar.")
+            return
+        await command_form(message)
+
     elif message_text.startswith("!skip"):
         if not active_voice_client or not active_voice_client.is_playing():
             await message.channel.send("Nothing is playing.")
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -633,6 +639,43 @@ def pick_random_emoji() -> str:
     return decoded_random_emoji
 
 
+async def command_form(message: Message):
+    # Escape strings so they can be assigned as literals within appscript
+    def escape_appscript(text: str):
+        return text.replace("\\", "\\\\").replace('"', '\\"')
+
+    # Constants in generated code, Mmke sure these strings are properly escaped
```
anoadragon453, line 649:
```suggestion
    # Constants in generated code. Make sure these strings are properly escaped
```

main.py
```diff
@@ -633,6 +639,43 @@ def pick_random_emoji() -> str:
...
+    high_string = "Masterpiece"
+    low_score = 0
+    high_score = 7
+
+    appscript = "function r(){"
+    # Setup and grab form
+    appscript += f'var f=FormApp.getActiveForm().setTitle("{default_title}");'
+    # Clear existing data on form
+    appscript += "f.getItems().forEach(i=>f.deleteItem(i));"
+    # Add new data to form TODO: properly escape double quotes and backslashes, etc
```
anoadragon453, line 661: What bugs does the current behaviour have?

main.py
```diff
@@ -35,6 +35,8 @@
 EMBED_DESCRIPTION_LIMIT = 4096
 # Max number of characters in an embed field.value
 EMBED_FIELD_VALUE_LIMIT = 1024
+# Max number of character is a normal Disord message
```
anoadragon453, line 38:
```suggestion
# Max number of characters in a normal Discord message
```

main.py
```diff
@@ -633,6 +641,52 @@ def pick_random_emoji() -> str:
     return decoded_random_emoji
 
 
+async def command_form(message: Message):
+    # Escape strings so they can be assigned as literals within appscript
+    def escape_appscript(text: str):
```
anoadragon453, line 646: Return types!

```suggestion
async def command_form(message: Message) -> None:
    # Escape strings so they can be assigned as literals within appscript
    def escape_appscript(text: str) -> str:
```

anoadragon453 (commented):
Looks pretty good for me, other than a couple wording fixes!

main.py
Cephian: None (that I know of), I did the TODO and forgot to remove it

PR #92, opened 2022-03-12
Further appscript golf

Cephian:
Small edit which makes generated Apps Script smaller when there are >= 5 songs. Instead of keeping the song ordinal ("1. ", "2. ", "3. ") as a literal string with the song title, capture the index of the `forEach` and add indices programmatically

### DIFF
--- a/main.py
+++ b/main.py
@@ -661,8 +661,7 @@ async def command_form(message: Message) -> None:
     # Add new data to form
     create_line = "[" + ",".join(
         [
-            '"{}. {}: {}"'.format(
-                index,
+            '"{}: {}"'.format(
                 escape_appscript(submit_message.author.display_name),
                 escape_appscript(song_format(local_filepath, attachment.filename)),
             )
@@ -672,7 +671,7 @@ async def command_form(message: Message) -> None:
             )
         ]
     )
-    create_line += '].forEach(s=>f.addScaleItem().setTitle(s).setBounds({},{}).setLabels("{}","{}"))'.format(
+    create_line += '].forEach((s,i)=>f.addScaleItem().setTitle(i+1+". "+s).setBounds({},{}).setLabels("{}","{}"))'.format(
         low_score, high_score, low_string, high_string
     )
     create_line += "}"


main.py
```diff
@@ -661,8 +661,7 @@ def escape_appscript(text: str) -> str:
     # Add new data to form
     create_line = "[" + ",".join(
         [
-            '"{}. {}: {}"'.format(
-                index,
+            '"{}: {}"'.format(
                 escape_appscript(submit_message.author.display_name),
                 escape_appscript(song_format(local_filepath, attachment.filename)),
             )
```
anoadragon453, line 667: `index` can now be removed from the next line.

main.py
```diff
@@ -672,7 +671,7 @@ def escape_appscript(text: str) -> str:
...
         ]
     )
-    create_line += '].forEach(s=>f.addScaleItem().setTitle(s).setBounds({},{}).setLabels("{}","{}"))'.format(
+    create_line += '].forEach((s,i)=>f.addScaleItem().setTitle(i+1+". "+s).setBounds({},{}).setLabels("{}","{}"))'.format(
```
anoadragon453, line 674: `${i+1}. ${s}` versus `i+1+". "+s` may be easier to read.

```suggestion
    create_line += '].forEach((s,i)=>f.addScaleItem().setTitle(`${i+1}. ${s}`).setBounds({},{}).setLabels("{}","{}"))'.format(
```

anoadragon453 (changes requested):

main.py
Cephian: I get an error running this, and it also doesn't seem like it would generate shorter Apps Script code, or that readability matters since we're trying to minimize character count. I also have hardly used JavaScript at all, so I'm not sure if there's some confusion with what's Python code vs what's Apps Script code or something.

anoadragon453: `${x}` within backticks are how you do parameter substitution in a string in js. But yes, it helps but readability though thats not the goal here :D

anoadragon453 (approved):

PR #93, opened 2022-03-15
Up the message scanning limit to 1000

anoadragon453:
Just in case we take a while to start, and people are talkative.

This doesn't appear to have any real performance impact.

### DIFF
--- a/main.py
+++ b/main.py
@@ -43,6 +43,8 @@ LIST_EMBED_COLOR = 0xDD2E44
 PLAY_EMBED_COLOR = 0x33B86B
 # The maximum character length of any song title or artist name
 MAXIMUM_SONG_METADATA_CHARACTERS = 1000
+# The maximum number of messages to scan for song submissions
+MAXIMUM_MESSAGES_TO_SCAN = 1000
 
 # SETTINGS
 # How many seconds to wait in-between songs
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -596,7 +596,7 @@ async def scrape_channel_media(
         os.mkdir(attachment_directory_filepath)
 
     # Iterate through each message in the channel
-    async for message in channel.history(limit=500, oldest_first=True):
+    async for message in channel.history(limit=1000, oldest_first=True):
```
Cephian, line 599: Maybe we should make this a constant?

In constants section:
```Python
# Maximum number of messages to scan for media attachments
MESSAGE_SCAN_LIMIT = 1000
```
On this line:
```suggestion
    async for message in channel.history(limit=MESSAGE_SCAN_LIMIT, oldest_first=True):
```

Cephian (changes requested):

main.py
Cephian: We have a comment in the constants section saying to see Discord docs for LIMIT variables, so if we do this we should either update that comment or change this variable name

anoadragon453: Good shout!

Cephian (approved):

PR #94, opened 2022-03-16
Show total length of songs after bust

### LINKED ISSUE #69: Add total sum of lengths of all played songs to "Thanks for coming"
We wouldn't want to let people know how long the stream is going to be from the start, because then it lets you guess the lengths of the last songs. However it's cool to know much content was created overall at the end.

Cephian:
Closes #69.

Sums the total lengths of all submissions to show at the end. This branch was tested in production at Busty's 6 (and it worked!)
image

The actual calculation is occurs when `!list` is called, but this doesn't increase list command latency since it is calculated after messages are sent and pinned.


Cephian:
I'm not totally sure why GitHub still says this branch has a merge conflict, but the "conflict" would be very easy to solve manually in the GitHub interface when merging, so consider this ready to merge.

anoadragon453:
@Cephian I fixed the conflict by doing:

```
git checkout total_len
git pull origin main
<fix conflicts>
git add -u && git commit
```

main.py
```diff
@@ -255,6 +257,21 @@ def song_format(
...
+        audio = MutagenFile(filename)
+        if audio:
+            return audio.info.length
+    except MutagenError as e:
+        # Ignore file and move on
+        print(f"Error reading length of {filename}:", e)
+        return None
+    except Exception as e:
+        print(f"Unknown error reading length of {filename}:", e)
+        return None
```
anoadragon453, line 272: `return None` could be extracted from both exception cases.

main.py
```diff
@@ -65,6 +65,8 @@
 original_bot_nickname: Optional[str] = None
 # Allow only one async routine to calculate !list at a time
 list_task_control_lock = asyncio.Lock()
+# Total length of all songs
```
anoadragon453, line 68:
```suggestion
# Total length of all songs in seconds
```

main.py
```diff
@@ -417,7 +451,10 @@ async def play_next_song(skip_count: int = 0) -> None:
 
         # Say our goodbyes
         embed_title = "❤️‍🔥 That's it everyone ❤️‍🔥"
-        embed_content = "Hope ya had a good **BUST!**"
+        embed_content = "*Total length of all submissions: {}*\n".format(
+            format_time(int(total_song_len))
+        )
```
anoadragon453, line 456: What do you think about putting the "total length" text at the bottom of the end message instead?

anoadragon453 (changes requested):
A few small changes, but overall looks pretty good!

main.py
```diff
@@ -255,6 +257,21 @@ def song_format(
     return content
 
 
+# Get length of a song
+def get_length(filename: str) -> Optional[float]:
+    try:
+        audio = MutagenFile(filename)
```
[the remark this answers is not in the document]
Cephian: If we do this it should be in a separate PR. I'd also want to see if it's doing things like storing the entire tag data (including all cover images) in the object which would increase our memory usage alot. At the very least, we are definitely making separate mutagen calls for both length and tag data when `!list`ing, and separate calls for both cover art and tag data when playing. which could probably be refactored and combined.

anoadragon453 (approved):
lgtm!

PR #98, opened 2022-03-19
Download files concurrently

Cephian:
I ran `!list` on a channel with all Busty's 4 songs with versions of Busty which:
1. Started all async downloads simultaneously and then waiting for them all to finish (this commit)
2. For each async download, started it and then `await`ed for it to finish before moving to the next download

I cleared the attachments folder in between runs of each version to avoid caching speedups. I performed this entire process about 5-6 times, each time comparing the time of the download phase of both (1) and (2).

While my testing approach was maybe slightly unscientific version (1) was *always* faster than (2) (usually ~9 seconds total download time vs ~12).

### DIFF
--- a/main.py
+++ b/main.py
@@ -645,7 +645,7 @@ async def scrape_channel_media(
             )
 
     # Save all files if not in cache
-    async def dl_file(attachment, attachment_filepath: str) -> None:
+    async def dl_file(attachment: Attachment, attachment_filepath: str) -> None:
         if not os.path.exists(attachment_filepath):
             await attachment.save(attachment_filepath)
 
--- a/main.py
+++ b/main.py
@@ -636,10 +636,6 @@ async def scrape_channel_media(
                 ),
             )
 
-            # Save file if not in cache
-            if not os.path.exists(attachment_filepath):
-                await attachment.save(attachment_filepath)
-
             channel_media_attachments.append(
                 (
                     message,
@@ -647,6 +643,14 @@ async def scrape_channel_media(
                     attachment_filepath,
                 )
             )
+
+    # Save all files if not in cache
+    async def dl_file(attachment, attachment_filepath: str) -> None:
+        if not os.path.exists(attachment_filepath):
+            await attachment.save(attachment_filepath)
+
+    tasks = [asyncio.create_task(dl_file(at, fp)) for _, at, fp in channel_media_attachments]
+    await asyncio.wait(tasks)
 
     # Clear unused files in attachment directory
     used_files = {path for (_, _, path) in channel_media_attachments}
--- a/main.py
+++ b/main.py
@@ -44,6 +44,8 @@ LIST_EMBED_COLOR = 0xDD2E44
 PLAY_EMBED_COLOR = 0x33B86B
 # The maximum character length of any song title or artist name
 MAXIMUM_SONG_METADATA_CHARACTERS = 1000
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -648,6 +644,16 @@ async def scrape_channel_media(
                 )
             )
 
+    # Save all files if not in cache
+    async def dl_file(attachment: Attachment, attachment_filepath: str) -> None:
+        if not os.path.exists(attachment_filepath):
+            await attachment.save(attachment_filepath)
+
+    tasks = [
+        asyncio.create_task(dl_file(at, fp)) for _, at, fp in channel_media_attachments
```
anoadragon453, line 653: Do we actually need to use `asyncio.create_task` here? [`asyncio.wait`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) states that it can act on any Awaitable object, which includes coroutines.

main.py
```diff
@@ -648,6 +644,16 @@ async def scrape_channel_media(
                 )
             )
 
+    # Save all files if not in cache
+    async def dl_file(attachment: Attachment, attachment_filepath: str) -> None:
```
anoadragon453, line 648: Have you tried running this on 40 songs? I wonder if running 40 downloads concurrently might lead to issues.

I looked into ways to do batches of tasks, though it's a little cumbersome.

Cephian: I tested with the 30 songs from busty's 4, though you're probably right that I should test with at least 60 or so before merging

Cephian: If we wanted to limit concurrent downloads I think the "correct" primitive to use would be a semaphore, which seems to not involve too much code to set up. See the first code sample [here](https://tutorialedge.net/python/concurrency/python-asyncio-semaphores-tutorial/).

main.py
Cephian: passing coroutines directly to `async.wait` is [deprecated](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) and will be removed in Python 3.11 (see italicized last few lines in `async.wait` section), so we should stick to creating tasks.

main.py
anoadragon453: That looks like a sane way of doing it, yes. And even if it works on your machine, it might be good to limit to ~10 songs anyhow. Our VPS has lots of bandwidth, but isn't massively powerful.

main.py
```diff
@@ -644,10 +646,14 @@ async def scrape_channel_media(
...
             )
 
+    download_semaphore = asyncio.Semaphore(value=MAXIMUM_CONCURRENT_DOWNLOADS)
+
     # Save all files if not in cache
     async def dl_file(attachment: Attachment, attachment_filepath: str) -> None:
+        await download_semaphore.acquire()
         if not os.path.exists(attachment_filepath):
             await attachment.save(attachment_filepath)
+        download_semaphore.release()
```
anoadragon453, line 720: We may want to use the `async with` syntax detailed [in the sephamore docs](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore), as it guarantees that the semaphore is released even if an Exception is raised?

Also we probably don't need to take a slot if the file does end up existing.

```suggestion
        if os.path.exists(attachment_filepath):
            return
        
        # Limit concurrent downloads
        await with download_semaphore:
            await attachment.save(attachment_filepath)
```

anoadragon453 (approved):

PR #103, opened 2022-03-30
Add volume multiplier to avoid clipping

### LINKED ISSUE #101: Lower audio output volume
Some users have complained that Busty's audio clips when she is set to 100% volume on Discord. This is probably an upstream Discord bug, and seems to occur on Windows but not Linux. To mitigate this we should turn Busty's output down 10% or so.

Cephian:
Closes #101. Adds a hard-coded volume multiplier. ~Currently marked as draft because I cannot find the optimal multiplier myself, as the clipping issue seems to only be reproducible on Windows. At some point I'll get in call with a Windows user to figure this out.~

Bobby and I found that the moment clipping becomes unnoticeable was setting the coefficient just over 0.5, so I set it to 0.5 since it seems like the type of number that whatever technical error is happening would actually need. Despite the volume multiplier being 0.5 down from 1, the audio is actually subjectively barely different in volume (some psychoacoustic log scaling something something).

### DIFF
--- a/main.py
+++ b/main.py
@@ -50,7 +50,7 @@ MAXIMUM_SONG_METADATA_CHARACTERS = 1000
 # The maximum number of messages to scan for song submissions
 MAXIMUM_MESSAGES_TO_SCAN = 1000
 # Volume multiplier to avoid clipping on Discord
-VOLUME_MULTIPLIER = 0.75
+VOLUME_MULTIPLIER = 0.5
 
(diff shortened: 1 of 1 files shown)


anoadragon453:
There seems to be an extra file named `=filter:a` that's snuck in.

anoadragon453 (approved):
dang, 50% is quite low. But I trust the science!

PR #104, opened 2022-03-31
Make voting required for all questions

Cephian:
Since we have a 0 voting option, we should make sure every vote is consciously chosen.


anoadragon453 (approved):
I'll admit, I used to also check over the entire survey a few times just to make sure I didn't miss one.

Good change!

PR #105, opened 2022-03-31
Title Google form based on bust # in channel name

Cephian:
Simple but effective. If any characters of the `current_channel` pass Python's `isdigit()` check, they are all concatenated together into a string _N_ and the voting form is titled "Busty's _N_ Voting". If there are no digits, the default title "Busty's Voting" is used.

### DIFF
--- a/main.py
+++ b/main.py
@@ -743,7 +743,7 @@ async def command_form(message: Message) -> None:
         return text.replace("\\", "\\\\").replace('"', '\\"')
 
     # Extract bust number from channel name
-    bust_number = "".join([c for c in message.channel.name if c.isdigit()])
+    bust_number = "".join([c for c in current_channel.name if c.isdigit()])
     if bust_number:
         bust_number = bust_number + " "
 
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -742,8 +742,13 @@ async def command_form(message: Message) -> None:
...
         return text.replace("\\", "\\\\").replace('"', '\\"')
 
+    # Extract bust number from channel name
+    bust_number = "".join([c for c in current_channel.name if c.isdigit()])
+    if bust_number:
+        bust_number = bust_number + " "
+
     # Constants in generated code, Make sure these strings are properly escaped
-    default_title = "Busty's Voting"
+    default_title = f"Busty's {bust_number}Voting"
```
anoadragon453, line 751: Should this still be called `default_title` if it ends up also being the real title?

anoadragon453 (approved):

main.py
Cephian: Good point, I'll rename it `form_title`

PR #110, opened 2022-05-04
Update screenshots

### LINKED ISSUE #109: Replace images in README.md
Given the fact that the Busty bot's profile picture has been updated and the reasons we updated it, we should change the promotional images to ones which use the new profile picture.

Cephian:
Closes #109. Updates screenshots to match new artwork. I went through every single "now playing" message and the one here was the one which used all features and had the least identifying information to cover up.


anoadragon453 (approved):
Hooray, new artwork!

Indeed looks to show off a good portion of the features. And good choice on the song :)

PR #112, opened 2022-06-14
Factor bust cleanup tasks to a separate function

anoadragon453:
Separated out from #78 in order to reduce the number of changes to reason about. The only change is that #78 originally added `cleanup` and `finish_bust` functions. However, the loop for cleaning up local files has these days been moved to when we scrape files, rather than end a bust. After removing that, the remaining `cleanup` function was quite short. So I inlined its contents into `finish_bust`.

As far as what behaviour this PR changes versus the current cleanup tasks of today; we now set `active_voice_client` and `original_bot_nickname` to `None`. #78 did this as well, and I think it's a reasonable change (else there's old bot state lying around).

Testing locally by letting a bust run its course and using `!stop` both worked as expected (bot's name originally restored, subsequent busts work fine).

### DIFF
--- a/main.py
+++ b/main.py
@@ -422,6 +422,40 @@ def command_skip() -> None:
     active_voice_client.stop()
 
 
+async def finish_bust() -> None:
+    """End the current bust."""
+    global active_voice_client
+    global original_bot_nickname
+    global current_channel_content
+    global current_channel
+    global total_song_len
+
+    # Disconnect from voice if necessary
+    if active_voice_client and active_voice_client.is_connected():
+        await active_voice_client.disconnect()
+
+    # Restore the bot's original guild nickname (if it had one)
+    bot_member = current_channel.guild.get_member(client.user.id)
+    if original_bot_nickname:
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -422,6 +422,40 @@ def command_skip() -> None:
...
+    global total_song_len
+
+    # Disconnect from voice if necessary
+    if active_voice_client and active_voice_client.is_connected():
+        await active_voice_client.disconnect()
+
+    # Restore the bot's original guild nickname (if it had one)
+    bot_member = current_channel.guild.get_member(client.user.id)
+    if original_bot_nickname:
+        await bot_member.edit(nick=original_bot_nickname)
```
Cephian, line 440: We will probably have to further factor this out if we fix #68, though I think it's fine keeping it in here for now. Up to you, since the purpose of this commit is largely code reorganization before other commits.

Cephian (approved):

main.py
anoadragon453: Yes, I think it's best to fix that separately, though hopefully such a change would be rather small.

PR #117, opened 2022-07-08
Send a message in the channel if song playback fails in some cases

### LINKED ISSUE #114: More gracefully handle an exception raised during playback
Currently if a song fails midway through playback, we'll just log an exception and carry on to the next song.

https://github.com/anoadragon453/busty/blob/012b6d257e722e6487ad727a25208b8772a3eb02/main.py#L556-L558

I think we could handle this in a slightly better way. Namely:

* Try replaying the song (though be careful of falling into an infinite loop).
* Send a message in the room that playback failed, rather than having no feedback on Discord. Otherwise everyone but the author will just think that the song simply ended abruptly.

anoadragon453:
Closes #114. We now stop the bust if an exception is raised by nextcord during playback, log the exception and message the channel about it (rather than playback simply stopping silently).

To users, this looks like:

image

Admins would then go check the traceback and decide whether the audio file is worth attempting to play again or not (or to skip and potentially try a re-encoded version at the end of the bust).

---

Note that I haven't dug too deeply into `nextcord.AudioPlayer` to see what exact exceptions may end up actually being passed to `ffmpeg_post_hook`, but this PR attempts to handle any all the same.

On top of exceptions raised by the above, I think it may be nice to handle *any* exception (even ones we raise) during a bust in the same manner, so I've factored the code out to a separate function. This PR doesn't attempt to catch any other exceptions yet though.

### DIFF
--- a/main.py
+++ b/main.py
@@ -116,9 +116,6 @@ async def on_ready() -> None:
 
 @client.event
 async def on_close() -> None:
-    # Unpin current "now playing" message if it exists
-    if now_playing_msg:
-        await try_set_pin(now_playing_msg, False)
     # Finish current bust (if exists) as if it were stopped
     await finish_bust(say_goodbye=False)
 
@@ -460,6 +457,7 @@ async def finish_bust(say_goodbye: bool = True) -> None:
     global current_channel_content
     global current_channel
     global total_song_len
+    global now_playing_msg
 
     # Disconnect from voice if necessary
     if active_voice_client and active_voice_client.is_connected():
@@ -469,6 +467,11 @@ async def finish_bust(say_goodbye: bool = True) -> None:
     if original_bot_nickname and current_channel:
         bot_member = current_channel.guild.get_member(client.user.id)
         await bot_member.edit(nick=original_bot_nickname)
+
+    # Unpin the currently playing song message if one is pinned
+    if now_playing_msg:
+        await try_set_pin(now_playing_msg, False)
+        now_playing_msg = None
 
     if say_goodbye:
         embed_title = "❤️‍🔥 That's it everyone ❤️‍🔥"
--- a/main.py
+++ b/main.py
@@ -45,6 +45,8 @@ MESSAGE_LIMIT = 2000
 LIST_EMBED_COLOR = 0xDD2E44
 # Color of "Now Playing" embed
 PLAY_EMBED_COLOR = 0x33B86B
+# Color of any error message embeds
+ERROR_EMBED_COLOR = 0xE22618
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -618,6 +627,38 @@ def ffmpeg_post_hook(e: BaseException = None):
...
+    Some info about the track (e.g. title) is included in the message sent
+    to the room to immediately inform admins about which song had issues, and
+    to give them the ability to assess whether the bust should be restarted from
+    that song, or the next one, after concluding that the song cannot be played.
+
+    Args:
+        e: The exception to log.
+        track_info: The title of the song.
+    """
+    global now_playing_msg
```
Cephian, line 645: Seems this we can remove this line, as `now_playing_msg` isn't mentioned further in the function

Cephian (commented):
Looking at this, I'm actually quite unsure how to test it. I changed the extension on a text file to .mp3 and the song just silently skipped in Discord, with only FFmpeg error messages printed in console
```
Error reading tags from file: attachments/995075428538667018.995075428320542880.mp3
Error reading tags from file: attachments/995075428538667018.995075428320542880.mp3
[mp3 @ 0x55d52997d640] Format mp3 detected only with low score of 1, misdetection possible!
[mp3 @ 0x55d52997d640] Failed to read frame size: Could not seek to 8305.
attachments/995075428538667018.995075428320542880.mp3: Invalid argument
```

From the documentation, [VoiceClient.play()](https://docs.nextcord.dev/en/stable/api.html#nextcord.VoiceClient.play) raises only
1. ClientException, when audio is already playing or not connected (shouldn't happen, not an issue with song, and everything would break anyways)
2. TypeError, which should be not possible since we directly pass in an FFmpegPCMAudio object
3. OpusNotLoaded, which I believe doesn't apply since we're using FFmpeg PCM and not Opus

The [FFmpegPCMAudio constructor](https://docs.nextcord.dev/en/stable/api.html#nextcord.ClientException) also seems to only raise ClientException if "The subprocess failed to be created". Since the process is created even for text files disguised as mp3, I'm guessing this only happens if FFmpeg is uninstalled or something similar.

In summary:
1) Do you know of a time where an exception can actually be thrown which is the "fault" of a song and not some large issue which would probably affect every single song?
2) We should see if we can catch FFmpeg errors on invalid files which don't seem to throw exceptions

anoadragon453:
Let's close this for now as I've mostly forgotten the state on it, and the diff will be different since the refactor.

I'm also not totally convinced anymore that it's a problem that really needs solving.

PR #119, opened 2022-08-17
Create the whole form automatically

anoadragon453:
This PR:

* Has the google appscript create (and name) the form itself, instead of needing to create one manually first.
* Adds the image itself. Use `!form link_to_gdrive_image` when using the form command.
* Prints links to the editable version of the form, and the public version, for easy sharing.
* Adds syntax highlighting to the code. This isn't necessary, but is nice if you need to quickly scan the golfed code for any reason.

After this PR, the workflow for creating a form is:

1. Pick an image from google drive (right click on the image->get link->copy link)
2. Run `!form https://drive.google.com/file/d/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/view?usp=sharing`
3. Copy/paste the code into [a new google appscript file](https://script.google.com/home)
4. Save and Run (Ctrl-S, Ctrl-R). Wait.
5. Copy/paste the published link (check it first) to Discord.

The execution log looks like:

image

Downsides:
* You need to move the form to the right folder afterwards. But this can be done outside of the form-creation rush.
    * We could theoretically automate this too (by adding the folder ID as an env var).
* You get a scarier warning about running the script as it now has the ability (via DriveApp) to create forms instead of just modifying existing ones. You can skip the warning though.
* Running the form is no longer idempotent. Running it multiple times will cause a new form to be created each time. This only results in clutter of your google drive though, which is not much of an issue.

### DIFF
--- a/main.py
+++ b/main.py
@@ -885,6 +885,13 @@ async def command_form(
     for i in range(0, len(appscript), chunk_size):
         await message.channel.send("```js\n{}```".format(appscript[i : i + chunk_size]))
 
+    # Tell the user how to generate the form
+    await message.channel.send(
+        "Copy/paste the above code into a new appscript project (replace anything already there). "
+        "Then click Save, and Run: https://script.google.com/home/projects/create\n\n"
+        "Authorize the project to use your Google Account when prompted. Click Advanced -> Go to ..."
+    )
+
 
 # Connect to Discord. YOUR_BOT_TOKEN_HERE must be replaced with
 # a valid Discord bot access token.
--- a/main.py
+++ b/main.py
@@ -835,8 +835,6 @@ async def command_form(
     appscript = "function r(){"
     # Setup and grab form
     appscript += f'var f=FormApp.create("{form_title}");'
-    # Clear existing data on form
-    appscript += "f.getItems().forEach(i=>f.deleteItem(i));"
     # Add questions to form
     appscript += "[" + ",".join(
         [
--- a/main.py
+++ b/main.py
@@ -813,7 +813,9 @@ def pick_random_emoji() -> str:
     return decoded_random_emoji
 
 
-async def command_form(message: Message, google_drive_image_link: Optional[str] = None) -> None:
+async def command_form(
+    message: Message, google_drive_image_link: Optional[str] = None
+) -> None:
     # Escape strings so they can be assigned as literals within appscript
     def escape_appscript(text: str) -> str:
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -858,7 +883,7 @@ def escape_appscript(text: str) -> str:
     # Print message in chunks respecting character limit
     chunk_size = MESSAGE_LIMIT - 6
     for i in range(0, len(appscript), chunk_size):
-        await message.channel.send("```{}```".format(appscript[i : i + chunk_size]))
+        await message.channel.send("```js\n{}```".format(appscript[i : i + chunk_size]))
```
anoadragon453, line 886: We could probably have busty post a link to https://script.google.com/home/projects/create as well for ease of use.

main.py
```diff
@@ -847,9 +854,27 @@ def escape_appscript(text: str) -> str:
         ")"
     )
 
+    if google_drive_image_link:
+        # Extract image file ID from the passed link
+        file_id_matches = re.match(
```
Cephian, line 859: It would be cool if we could also
1) pass a non g-drive image using URLFetchApp ([example](https://developers.google.com/apps-script/reference/forms/image-item))
2) pass image via an attachment to the `!form` command

Though that seems like work for a future commit :^)

Cephian (changes requested):
We should update the README with the new `!form` command option.

main.py
anoadragon453: I ended up doing this in https://github.com/anoadragon453/busty/pull/119/commits/ca3ffaf19e1847242aabaace193b796d1ea82f98. It looks like the following:

image

main.py
anoadragon453: Indeed, that'd be very fancy.

main.py
```diff
@@ -847,18 +854,43 @@ def escape_appscript(text: str) -> str:
...
+            appscript += (
+                f';f.addImageItem().setImage(DriveApp.getFileById("{file_id}"))'
+            )
+
     # Add comments/suggestions to form
     appscript += ";f.addParagraphTextItem().setTitle('Comments/suggestions')"
 
+    # Print links to the form
+    appscript += (
+        ';console.log("Edit: "+f.getEditUrl()+"\\n\\nPublished: "+f.getPublishedUrl());'
```
anoadragon453, line 875: I wish these could be clickable links, but as far as I can see you can't get clickable links in the app script execution log.

Cephian (approved):

PR #125, opened 2022-08-20
Split main.py functionalities into separate files

### LINKED ISSUE #121: Split the code into multiple files
Right now Busty's `main.py` is a monolithic 900+ lines of code, with many only somewhat related parts being interspersed with each other. We should split related functions into separate modules for better organization

Cephian:
Fixes #121. Since massive changes like this will totally mess up diffs, I tried to make as few logic changes as possible when copying and pasting. This change has been tested.

The only real "logic change" I did was to remove the `get_random_emoji()` function and get a random emoji inline, since we only used it once and it was the only utility function which fit in neither `song_utils` nor `discord_utils`. I put `command_form` into a separate file `voting.py` for now, since eventually i'd like all of the forms logic to be there and have only thin wrapper code translating message --> command in either `main.py` or `bust.py`


anoadragon453 (approved):
This definitely looks like a great step in the right direction! I assume you've tested it :)

just a tiny correction below.

anoadragon453:
Note that since we're moving essentially all code, the result is that any invocation of `git blame` will likely run into this commit, causing noise. We can remove this by creating a `.git-blame-ignore-revs` file at the root of the repository, containing something like the following:

```
# Commits in this file will be removed from GitHub blame results.
#
# To use this file locally, use:
#   git blame --ignore-revs-file="path/to/.git-blame-ignore-revs" <files>
#
# or configure the `blame.ignoreRevsFile` option in your git config.
#
# If ignoring a pull request that was not squash merged, only the merge
# commit needs to be put here. Child commits will be resolved from it.

# Split main.py functionalities into separate files (https://github.com/anoadragon453/busty/pull/125).
<merge commit>
```

we'll need to merge first before we know what commit to put though.

Issue #122, opened 2022-08-20
Refactor to avoid heavy usage of optional global variables


Cephian:
Busty's uses a lot of Optional global variables which are only not `None` when the bust is active or items are listed. This makes static type checking very difficult, doesn't couple related components  very well, and is just plain ugly. We should put all of these bust related variables into a `BustController` class, which can better keep track of bust state.

PR #127, opened 2022-08-21
Create BustController

Cephian:
The BustController basically doesn't do anything in this commit, and is only setting the stage for #122 by getting unreadable diffs out of the way before making logic changes. Most lines in `bust.py` have been moved to a one-deeper level of indent.

The only real organizational change I made other than throwing (almost) everything in bust.py into a BustController class is rename the `command_xyz()` methods to just `xyz()`. This commit has been tested.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -118,9 +118,9 @@ async def on_message(message: Message) -> None:
         # Pull the google drive link to the form image from the message (if it exists)
         command_args = message.content.split()[1:]
         if command_args:
-            await voting.command_form(message, google_drive_image_link=command_args[0])
+            await voting.form(message, google_drive_image_link=command_args[0])
         else:
-            await voting.command_form(message)
+            await voting.form(message)
 
     elif message_text.startswith("!skip"):
         if not bust.active_voice_client or not bust.active_voice_client.is_playing():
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -103,40 +107,40 @@ async def on_message(message: Message) -> None:
...
             return
 
         # Pull the google drive link to the form image from the message (if it exists)
         command_args = message.content.split()[1:]
         if command_args:
-            await voting.command_form(message, google_drive_image_link=command_args[0])
+            await voting.form(message, google_drive_image_link=command_args[0])
         else:
-            await voting.command_form(message)
+            await voting.form(message)
```
anoadragon453, line 127: I might give these methods slightly more informative names while we're changing the names anyhow. Such as:

* `generate_form`
* `skip_song`
* `send_song_list`

Since they're no longer directly tied to the command names.

anoadragon453 (approved):
LGTM otherwise!

great to see those `global` keywords disappear.

src/main.py
Cephian: I changed form() and skip(). Not changed list since that is waiting to be factored out in #130 anyways.

PR #129, opened 2022-08-21
Fix emoji display

Cephian:
It turns out the single "logic change" in #125 had a bug I didn't notice, it was using emoji names instead of emoji themselves.

image


Cephian:
I'm just going to push this myself

Issue #123, opened 2022-08-20
Change control flow of playing next song


Cephian:
As it is now, the current song is advanced by spawning a new coroutine and actually altering the list of all songs, destroying data. It makes more sense to store song metadata immutably over the course of the bust, and instead just iterate over this list, spawning a new process each to play it/wait/etc.

We will probably have to use some sort of thread join or lock in order to spawn new processes. The reason I suspect we even need to spawn new processes at all is so that they can be properly interrupted by `!skip`, `!stop`, and `!next`.

PR #130, opened 2022-08-21
Decouple BustController

Cephian:
Taking some steps to move towards a more modular BustController. BustController should only be instantiated via a new `bust.create_controller()` method, which either constructs a valid controller or returns None. Bust state can be checked externally with the `BustController.active()` function.

Three optional variables still remain in BustController, but for now these all make sense as optional as they move in and out of existence throughout the BustController's lifecycle.

The code which calculates total song length has been moved to the BustController constructor, whereas previously it was in the list command (which has largely become the create_controller function). A few other things like whether there is an active lock are now checked within create_controller as well.

Management of active controllers is done via a map from guild id -> controller called `controllers` in `main.py`. Management is a bit suboptimal right now since we have no real way of making a "callback" when a bust finishes to remove itself from the map. This should be fixed after #123 if we write it such that `BustController.play()` is an async function which returns when the bust is over.

Despite it seeming like this setup supports multiple servers, since everything still shares an attachment folder this is untrue. It would not be too hard to have an attachment folder per-BustController in the future however.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -21,18 +21,21 @@ import config
 import discord_utils
 import song_utils
 
+
 # Allow only one async routine to calculate !list at a time
 list_task_control_lock = asyncio.Lock()
 
 
 class BustController:
-    def __init__(self, client: Client):
+    def __init__(
+        self,
+        client: Client,
+        current_channel_content: List[Tuple[Message, Attachment, str]],
+        current_channel: TextChannel,
+    ):
 
         # The actively connected voice client
-        self.active_voice_client: Optional[VoiceClient] = None
-        # The nickname of the bot. We need to store it as it will be
-        # changed while songs are being played.
-        self.original_bot_nickname: str = None
+        self.voice_client: Optional[VoiceClient] = None
         # Currently pinned "now playing" message ID
         self.now_playing_msg: Optional[Message] = None
         # Keep track of the coroutine used to play the next track, which is active while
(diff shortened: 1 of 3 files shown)


src/voting.py
```diff
@@ -3,24 +3,24 @@
...
+from bust import BustController
 
 # TODO: Eventually the function in this file should be rewritten not interact with
 # Discord at all, and instead called by a wrapper in bust.py with most
 # internal logic remaining here
 
 
-async def command_form(
-    message: Message, google_drive_image_link: Optional[str] = None
+async def form(
```
anoadragon453, line 15: I might give these methods slightly more informative names while we're here. Such as:

* `generate_form`
* `skip_song`
* `send_song_list`

Since they're no longer directly tied to the command names.

src/bust.py
```diff
@@ -26,39 +26,61 @@
...
         # Currently pinned "now playing" message ID
         self.now_playing_msg: Optional[Message] = None
         # Keep track of the coroutine used to play the next track, which is active while
         # waiting in the intermission between songs. This task should be interrupted if
         # stopping playback is requested.
         self.play_next_task: Optional[asyncio.Task] = None
 
+        # The nickname of the bot. We need to store it as it will be
+        # changed while songs are being played.
+        self.original_bot_nickname: str = ""
```
anoadragon453, line 47: `None` is actually [a valid state](https://docs.nextcord.dev/en/stable/api.html?highlight=edit#nextcord.Member.nick) for a guild member's nick - it just means use the username (i.e. no nick). A bot's nick may be `None` pre-bust.

So I think the type of original_bot_nickname` should remain an `Optional[str]`.

src/main.py
```diff
@@ -32,15 +31,15 @@ async def on_ready() -> None:
...
-    # Finish current bust (if exists) as if it were stopped
-    await bc.finish(say_goodbye=False)
+    # Finish all running busts on close
+    for _, bc in controllers.items():
+        await bc.finish(say_goodbye=False)
 
 
 @client.event
 async def on_message(message: Message) -> None:
+    global controllers
```
anoadragon453, line 41: Not for this PR, but another thing we could do to get rid of `global`s is to wrap most of the `main.py` login in a `class BustyBot`, which is then instantiated and has, say, a `run` function.

`controllers` would then just be a variable under the `BustyBot` class. So:

```py
class BustyBot:
    controllers: Dict[int, BustController] = {}

    def __init__(self, client: Client) -> "BustController":
        self.client = client

        # Connect to discord
        self.client.run()

    @client.event
    async def on_message(self, message: Message) -> None:
        ...
```

anoadragon453 (changes requested):
Great improvements overall - it's starting to really take shape!

Some gotchas, suggestions and opinions:tm: below.

Cephian:
Thanks for the suggestions, I fixed all the ones you mentioned (except the BustyBot class comment) and tested the bot again to make sure nothing is horribly broken

anoadragon453 (approved):

PR #131, opened 2022-08-21
Move Discord token to config.py

Cephian:
So we read all env variables in the same place


src/config.py
```diff
@@ -28,6 +28,8 @@
 attachment_directory_filepath = os.environ.get("BUSTY_ATTACHMENT_DIR", "attachments")
 # The Discord role needed to perform bot commands
 dj_role_name = os.environ.get("BUSTY_DJ_ROLE", "bangermeister")
+# The Discord bot token to use
+discord_token = os.environ.get("BUSTY_DISCORD_TOKEN", None)
```
anoadragon453, line 32:
```suggestion
discord_token = os.environ.get("BUSTY_DISCORD_TOKEN")
```

The second argument of `dict.get` defaults to `None`.

anoadragon453 (approved):
lgtm otherwise!

PR #135, opened 2022-09-21
Generate voting form directly

### LINKED ISSUE #12: Autogeneration of Google Forms for Voting

Cephian:
Closes #12. Automatically generate and send voting form on `!list`.

Also added an `!image` to load an image to be placed at the front of the voting form. A loaded image does not get unset automatically when a form is generated, but it can be cleared manually with `!image clear`. To set an image use either `!image <url>` or run `!image` with an image attachment.

Also added another option `full` to `!list`. When either `full` is passed to `!list` or the channel the command was issued in is the same as the target channel, the tracklist will be pinned and the form will be generated (and pinned). This is so form generated can be tested before the show.

Check the diffs on `README.md` for other user-facing changes.

I also added the :+1: and :stop_sign: emoji to `config.py` instead of locally to `bust.py` since they are also used in `!image`.

image

**KNOWN ISSUES**
1. When `!image` is used, the attachment/url is not scanned at all for validity until the Form is generated. The Form is generated such that if the image is invalid, then generation will still work but without the image.
2. There is no error/warning when the user-passed `BUSTY_GOOGLE_FORM_FOLDER` is invalid. The move command just silently fails, though the form is still generated.
3. If the bot is restarted, the loaded image is lost (unlike the `!list` cache). Would be nice to save state to disk when set so that we can set the bust image at any point and restart/update busty in the meantime.
4. While this is MUCH faster than Apps Script, it still takes 5-6 seconds to generate the form. This is done sequentially after list generation, but in theory most steps of Form generation can be done before knowing the track-list (about 2/3 of the total time). We could in theory do those steps simultaneously with track downloading using this async Google API library https://github.com/omarryhan/aiogoogle. Probably not worth it to be honest. A good intern project? ;^)
5. Not so much an issue, but I'm not sure if we should also keep `!form` around "just in case" for now.

I'm fine merging with these issues, but 1 & 2 & 3 should have bugs created on merging if this is the case.


src/main.py
```diff
@@ -112,19 +115,35 @@ async def on_message(message: Message) -> None:
...
+                    f"Loaded image: {loaded_image}\n\nTo change this image, "
+                    "either run `!image <image_url>` or"
+                    " just `!image` with a valid media attachment. "
+                    "To clear this image, run `!image clear`"
+                )
+            else:
+                message_reply_content = (
+                    "No image is currently loaded.\n\nTo add an image, "
+                    "either run `!image <image_url>` or"
+                    " just `!image` with a valid media attachment.\n"
```
anoadragon453, line 144: Best to be consistent on where the space is put for multi-line strings. Either at the beginning or the end of each string. Here they're mixed.

src/bust.py
```diff
@@ -287,17 +288,44 @@ def ffmpeg_post_hook(e: BaseException = None):
...
+    client: Client, message: Message, image_url: Optional[str] = None
 ) -> Optional[BustController]:
     """Attempt to create a BustController given a channel list command"""
-    command_success = "\N{THUMBS UP SIGN}"
-    command_fail = "\N{OCTAGONAL SIGN}"
 
     # Ensure two scrapes aren't making/deleting files at the same time
     if list_task_control_lock.locked():
-        await message.add_reaction(command_fail)
+        await message.add_reaction(config.COMMAND_SUCCESS)
```
anoadragon453, line 328: Was the change to success here on purpose?

src/bust.py
```diff
@@ -387,10 +417,27 @@ async def create_controller(
...
                 await discord_utils.try_set_pin(list_message, True)
+            # Wrap form generation in try/catch so we don't block !list if it fails
+            form_url = None
+            try:
+                form_url = await bc.get_google_form_url(image_url)
+            except Exception as e:
+                print("Unknown error generating form", e)
+
+            if form_url is not None:
+                vote_emoji = ":ballot_box:"
```
anoadragon453, line 436: To be consistent with the definitions in `config.py`, this can also be `"\N{BALLOT BOX WITH BALLOT}"`.

src/main.py
```diff
@@ -23,6 +22,9 @@
 
 controllers: Dict[int, BustController] = {}
 
+# Cached image to use for next bust
+loaded_image: Optional[str] = None
```
anoadragon453, line 26: Hmm, if ever wanted to make concurrent busts a thing this would need to be scoped to a server. But for now I think it's fine.

anoadragon453 (changes requested):
> Not so much an issue, but I'm not sure if we should also keep `!form` around "just in case" for now.

Mmm, with the ability to test the form beforehand I'd be surprised if we needed to regenerate after running `!list` publicly. Even then we could just re-run `!list` (and delete any audio files that were posted in the meantime).

README.md
```diff
@@ -54,10 +54,17 @@ servers maximum, unless you verify your bot with Discord.
 Copy the bot token, and ensure that the environment variable `BUSTY_DISCORD_TOKEN` contains
 the bot token when running the bot.
 
+If you'd like automatic Google Form generation, also ensure a Google service account key file
```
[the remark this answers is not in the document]
Cephian: I honestly don't remember what I did, I think it was a number of steps. Probably a link is good enough, a full tutorial might be overkill

README.md
```diff
@@ -92,11 +99,11 @@ The expected flow for running a bust is:
 
 ### Command Reference
 
-1. `!list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`.
+1. `!list [<channel>] [all]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`. If the `all` flag is set or if the channel listed is the channel the command was sent in, list messages are pinned and a Google Form for voting is generated
```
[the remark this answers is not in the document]
Cephian: I thought about this, but it felt to me like more of a Unix thing than Discord. I'm fine with it as is, or I'm fine changing it if that's what you'd prefer

src/main.py
Cephian: Yes, though they're already but possible because the attachments directory will remove anything listed from another server on `!list`. So we can change this if we ever want to, but it's fine for now imo

src/forms.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Move form to correct folder + rename
+    file = (
+        drive_service.files()
+        .get(fileId=form_id, fields="capabilities/canMoveItemWithinDrive, parents")
+        .execute()
+    )
+    if file["capabilities"]["canMoveItemWithinDrive"]:
+        file_parent_id = file["parents"][0]
+        # TODO: Command fails silently if config.google_form_folder is invalid
```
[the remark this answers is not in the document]
Cephian: I probably can, my guess is that it would add another second of query time. I'm guessing actually the return value of this function will tell me if an error occurred, so I'll look into that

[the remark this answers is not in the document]
Cephian: It now prints an error by catching an exception. I'm not actually sure why the exception didn't just dump to screen when I tried this before.

anoadragon453 (approved):
Sorry for the slow turn-around times. This LGTM!

PR #141, opened 2022-11-09
Make loaded image persistent across reboots

Cephian:
Added a new class PersistentString, which keeps the state of an `Optional[str]` variable in a passed filepath. If the file exists when an instance is created, the string contents are loaded from file. Otherwise, they are none. Whenever the string is set, the file is either written to or deleted depending on whether the string is `None` or not.

Made `loaded_image` a PersistentString, with filepath pointed to by either env var `BUSTY_IMAGE_STATE_FILE`, or with a default value of `.image_state`.

There are a number of different ways this could have been engineered, lmk if you can think of a nicer one and we should do that. This solution seems simple and small enough that it wouldn't be too hard to replace it with something more generic and robust later if we want to start keeping other persistent data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ The complete list of environment variable configuration options is:
 1. `BUSTY_ATTACHMENT_DIR` - Directory to save attachments (default = attachments)
 1. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
 1. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The Python module to import containing the emoji list (default = emoji_list)
+1. `BUSTY_IMAGE_STATE_FILE` - The location of the file which saves the state of the currently loaded image (default = .image_state)
 
 A random emoji is displayed for each song played during a bust. The list of possible
 emoji is defined in [emoji_list.py](src/emoji_list.py). If you would like to customize
--- a/src/persistent.py
+++ b/src/persistent.py
@@ -18,7 +18,7 @@ class PersistentString:
         self._value: Optional[str] = None
         # Load string if it exists
         with contextlib.suppress(FileNotFoundError):
-            with open(self.filepath) as f:
+            with open(self.filepath, "r") as f:
                 self._value = f.read()
 
     def set(self, value: Optional[str]):
@@ -26,11 +26,11 @@ class PersistentString:
 
         # Either save string, or delete state file if None
         if self._value is None:
-            with open(self.filepath, "w") as f:
-                f.write(self._value)
-        else:
             with contextlib.suppress(FileNotFoundError):
                 os.remove(self.filepath)
+        else:
+            with open(self.filepath, "w") as f:
+                f.write(self._value)
 
     def get(self) -> Optional[str]:
         return self._value
--- a/src/persistent.py
+++ b/src/persistent.py
@@ -1,0 +1,36 @@
+import contextlib
+import os
+from typing import Optional
+
+
+class PersistentString:
+    """A string wrapper which is persistent across reboots.
+
+    Methods:
+        get(value):
+            Gets the value of the string
+        set(value):
+            Sets the value of the string and writes to disk
+    """
+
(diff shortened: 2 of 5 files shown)


src/persistent.py
```diff
@@ -0,0 +1,36 @@
...
+        # Load string if it exists
+        with contextlib.suppress(FileNotFoundError):
+            with open(self.filepath) as f:
+                self._value = f.read()
+
+    def set(self, value: Optional[str]):
+        self._value = value
+
+        # Either save string, or delete state file if None
+        if self._value is None:
```
anoadragon453, line 28: I think this conditional might be backwards?

src/persistent.py
```diff
@@ -0,0 +1,36 @@
...
+            with open(self.filepath, "w") as f:
+                f.write(self._value)
+        else:
+            with contextlib.suppress(FileNotFoundError):
```
anoadragon453, line 32: A slightly more readable way to handle this could be to use `if os.exists(...)`.

src/persistent.py
```diff
@@ -0,0 +1,36 @@
...
+    """A string wrapper which is persistent across reboots.
+
+    Methods:
+        get(value):
+            Gets the value of the string
+        set(value):
+            Sets the value of the string and writes to disk
+    """
+
+    def __init__(self, filepath: str):
```
anoadragon453, line 16:
```suggestion
    def __init__(self, filepath: str) -> None:
```

To be consistent with the rest of the codebase.

README.md
```diff
@@ -69,6 +69,7 @@ The complete list of environment variable configuration options is:
 1. `BUSTY_ATTACHMENT_DIR` - Directory to save attachments (default = attachments)
 1. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
 1. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The Python module to import containing the emoji list (default = emoji_list)
+1. `BUSTY_IMAGE_STATE_FILE` - The location of the file which saves the state of the currently loaded image (default = .image_state)
```
anoadragon453, line 72: Starting to wonder if we should just have a "data" folder that this file and an "attachments" subfolder could live in, rather than have an env var for each. What do you think?

anoadragon453 (changes requested):
I think this implementation would work well, and we don't need to get too complex by actually inheriting from `str` or overriding dunder methods. I do think might be starting to get into the realm of wanting a database, or at least a .json file to keep persistent bot state. Nothing else immediately comes to mind to store though, so this is fine for now imo

README.md
Cephian: I like this idea, I say we do it in a separate commit after this one to keep commits small and single-purpose

src/persistent.py
Cephian: `os.exists(...)` is _technically_ less safe because the file could be removed in between the check and when the read is finished :^)

If you think we should replace it with a try/except bc nobody knows what contextlib is then I can understand that though lol. I didn't do that because it's longer and a bit uglier

src/persistent.py
Cephian: not sure how that happened

src/persistent.py
anoadragon453: While I think that case is pretty unlikely, I admire the defensive programming :)

But I do think a try/except would be more readable. Higher readability often comes at the cost of more lines/lower performance. But it's often worth it if those things aren't mission-critical.

anoadragon453 (approved):
LGTM, other than a try/except at https://github.com/anoadragon453/busty/blob/ad809046e36a74ce012539484fe9c76d11f1e1b5/src/persistent.py#L20 and https://github.com/anoadragon453/busty/blob/ad809046e36a74ce012539484fe9c76d11f1e1b5/src/persistent.py#L29

src/persistent.py
Cephian: Sure, changed it

PR #142, opened 2022-11-27
Refactor some constants to constants.py

### LINKED ISSUE #128: Various refactor-blocked changes
Not doing this right now because a number of large code changes are currently happening where minimized lines changed is key.

1) The hardcoded 32 nickname limit should  be moved to `config.py` as a constant.
2) The hardcoded 8MB image data limit in `song_util.py` should also be a constant.

Cephian:
Closes #128, moved two constants to `constants.py`.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -10,7 +10,7 @@ EMBED_FIELD_VALUE_LIMIT = 1024
 MESSAGE_LIMIT = 2000
 # Max number of characters in a Discord nickname
 NICKNAME_CHAR_LIMIT = 32
-# (Lower bound for) Max allowed bytes in a Discord attachment
+# (A safe lower bound for) max allowed bytes in a Discord attachment
 ATTACHMENT_BYTE_LIMIT = 8000000
 # Color of !list embed
 LIST_EMBED_COLOR = 0xDD2E44
--- a/src/config.py
+++ b/src/config.py
@@ -8,6 +8,10 @@ EMBED_DESCRIPTION_LIMIT = 4096
 EMBED_FIELD_VALUE_LIMIT = 1024
 # Max number of characters in a normal Disord message
 MESSAGE_LIMIT = 2000
+# Max number of characters in a Discord nickname
+NICKNAME_CHAR_LIMIT = 32
+# (Lower bound for) Max allowed bytes in a Discord attachment
+ATTACHMENT_BYTE_LIMIT = 8000000
 # Color of !list embed
 LIST_EMBED_COLOR = 0xDD2E44
 # Color of "Now Playing" embed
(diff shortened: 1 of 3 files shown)


src/config.py
```diff
@@ -8,6 +8,10 @@
 EMBED_FIELD_VALUE_LIMIT = 1024
 # Max number of characters in a normal Disord message
 MESSAGE_LIMIT = 2000
+# Max number of characters in a Discord nickname
+NICKNAME_CHAR_LIMIT = 32
+# (Lower bound for) Max allowed bytes in a Discord attachment
+ATTACHMENT_BYTE_LIMIT = 8000000
```
anoadragon453, line 14: Starting to think we should move these out to a `constants.py` file. Seeing `config.NICKNAME_CHAR_LIMIT` makes it seem that the character limit is configurable when it's not the case.

---

Edit: I'm just now seeing the title of the PR and realising we might be on the same page lol

src/song_utils.py
```diff
@@ -162,7 +162,7 @@ def get_cover_art(filename: str) -> Optional[File]:
 
     # Make sure it doesn't go over 8MB
     # This is a safe lower bound on the Discord upload limit of 8MiB
```
anoadragon453, line 164:
```suggestion
    # Make sure it doesn't go over the maximum size allowed for a Discord attachment.
    # Important for when the bot later posts the image during a bust.
```

I'd move the 8MB-specific text to the definition of `ATTACHMENT_BYTE_LIMIT`.

anoadragon453 (changes requested):

src/config.py
Cephian: Haha yeah wow didn't notice I did that. I agree, though since it might affect a lot of files and I have some other PRs I want to do soon that might also affect a lot of files and I consider "higher priority" I think we should merge this for now and make an issue about separating config and constants

Cephian: Actually, we could just rename `config.py` to `constants.py` or `settings.py` or something else since they are effectively fixed at runtime. And some of the discord-pulled constants are in theory something you could WANT to change to something that's not the exact limit imposed by Discord.

I think they have enough semantic similarity and the file is short enough that there's not really an organizational reason to separate them.

anoadragon453: I do still think of constants as things that are hardcoded, rather than configurable values. `settings.py` sounds good.

src/config.py
```diff
@@ -8,6 +8,10 @@
 EMBED_FIELD_VALUE_LIMIT = 1024
 # Max number of characters in a normal Disord message
 MESSAGE_LIMIT = 2000
+# Max number of characters in a Discord nickname
+NICKNAME_CHAR_LIMIT = 32
+# (Lower bound for) Max allowed bytes in a Discord attachment
```
anoadragon453, line 13:
```suggestion
# (A safe lower bound for) max allowed bytes in a Discord attachment
```

anoadragon453 (approved):
Had totally missed you'd already mentioned the lower bound bit in `config.py`. I've updated it here to be slightly clearer to me, but feel free to take it or leave it.

PR #146, opened 2022-11-28
Convert to slash commands

### LINKED ISSUE #143: Convert to slash commands
Move to the new Discord model of slash commands with more rich interactions

Cephian:
🚀 〜THE FUTURE IS HERE〜 🚀

Resolves #143. Commands have been converted to slash commands. See the shiny collage down below. Some notes

1) Since slash command arguments are typed, balancing the "full" and the "target channel" options for /list  became annoying, so I removed the "full" option (which was previously easy to throw in). We can implement it in the future if we ever find we need it, but it's largely not useful for testing so far.
2) Since slash command arguments are typed, the single `!image` command had to be split into four subcommands, `/image view`, `/image upload`, `/image url`, `/image clear`.
3) The use case of "acknowledgement reactions" has been replaced by both (a) a "thinking" response" and (b) the fact that commands issued disappear, so we need to leave a full message in history anyways. As a sort of corollary the command success and command fail emoji have been removed from `config.py`.
4) Slash commands have this weird thing where they all show to all server members by default and you have to do visibility permissions yourself in server settings. Unprivileged command access is  handled ofc, I just thought this was weird.
5) (Global) Slash commands do this other weird thing where they take an hour to register when you first run your bot. For testing purposes this can be circumvented by explicitly specifying a list of servers for which they should be updated instantly. I've added an env var option to add one of these testing servers.

image
image
image
image

P.S. Bot permissions are confusing, running instances will probably have to be re-invited to servers. I believe this is a minimal set of necessary perms
https://cdn.discordapp.com/attachments/1046363901425635399/1046581622574682212/image.png

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ The expected flow for running a bust is:
 
 ### Command Reference
 
-1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`. 
+1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `/bust`. 
 1. `/bust [<song #>]` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work. Specifying a song index will skip to that index before playing.
 1. `/image upload [<attachment>]` - Upload an image to be used in the Google Form generated when running `/list`.
 1. `/image url [<url>]` - Queue an image to be used in the Google Form generated when running `/list`.
--- a/README.md
+++ b/README.md
@@ -110,7 +110,9 @@ The expected flow for running a bust is:
 1. `/stop` - Stop busting early :scream: :scream: :scream: 
 
 Users must have the `bangermeister` role to use commands by default, though this role can
-be modified by passing the `BUSTY_DJ_ROLE` environment variable.
+be modified by passing the `BUSTY_DJ_ROLE` environment variable. It is highly recommended that
+you disable command visibility for those without permissions to run them in the Integrations
+section of your server's settings.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -100,12 +100,14 @@ The expected flow for running a bust is:
 
 ### Command Reference
 
-1. `!list [<channel>] [full]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`. 
-If the `full` flag is set or if the channel listed is the channel the command was sent in, list messages are pinned and a Google Form for voting is generated.
-1. `!bust [<song #>]` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work. Specifying a song index will skip to that index before playing.
-1. `!image [<url>]` - Queue an image to be used in the Google Form generated when running `!list`. Run `!image` with no arguments for full usage information.
-1. `!skip` - Skips the current track :scream: 
-1. `!stop` - Stop busting early :scream: :scream: :scream: 
+1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `!bust`. 
+1. `/bust [<song #>]` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work. Specifying a song index will skip to that index before playing.
+1. `/image upload [<attachment>]` - Upload an image to be used in the Google Form generated when running `/list`.
+1. `/image url [<url>]` - Queue an image to be used in the Google Form generated when running `/list`.
+1. `/image view [<url>]` - View currently loaded image to be used in the Google Form generated when running `/list`.
+1. `/image clear [<url>]` - Clear currently loaded image to be used in the Google Form generated when running `/list`.
+1. `/skip` - Skips the current track :scream: 
+1. `/stop` - Stop busting early :scream: :scream: :scream: 
 
 Users must have the `bangermeister` role to use commands by default, though this role can
 be modified by passing the `BUSTY_DJ_ROLE` environment variable.
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ The complete list of environment variable configuration options is:
 1. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
 1. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The Python module to import containing the emoji list (default = emoji_list)
 1. `BUSTY_IMAGE_STATE_FILE` - The location of the file which saves the state of the currently loaded image (default = .image_state)
-1. `BUSTY_TESTING_SERVER_ID` - For developers only. Specify a testing server id to avoid 1 hour command update delay (see [this Discord API issue](https://github.com/discord/discord-api-docs/issues/2372#issuecomment-761161082) for details) (default = None)
+1. `BUSTY_TESTING_GUILD_ID` - For developers only. Specify a testing guild id to avoid 1 hour command update delay (see [this Discord API issue](https://github.com/discord/discord-api-docs/issues/2372#issuecomment-761161082) for details) (default = None)
 
 A random emoji is displayed for each song played during a bust. The list of possible
 emoji is defined in [emoji_list.py](src/emoji_list.py). If you would like to customize
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ The complete list of environment variable configuration options is:
 1. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
 1. `BUSTY_CUSTOM_EMOJI_FILEPATH` - The Python module to import containing the emoji list (default = emoji_list)
 1. `BUSTY_IMAGE_STATE_FILE` - The location of the file which saves the state of the currently loaded image (default = .image_state)
-1. `BUSTY_TESTING_SERVER_ID` - For developers only. Specify a testing server id to avoid 1 hour command update delay (default = None)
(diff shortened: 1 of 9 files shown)


src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
-    if message.author == client.user:
+# List command
+@client.slash_command()
+@application_checks.has_role(config.dj_role_name)
+async def list(
+    interaction: Interaction,
+    list_channel: Optional[TextChannel] = SlashOption(required=False),
+) -> None:
+    """Download and list all media sent in a chosen text channel."""
+    bc = get_controller(interaction.guild_id)
```
anoadragon453, line 71: Note that `interaction.guild_id` will be `None` here (and on other commands) if this interaction came in via a DM.

We probably want to just disable slash commands via DM? (via the [`dm_permission`](https://docs.nextcord.dev/en/stable/api.html#nextcord.Client.slash_command) argument of `@slash_command`).

src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
-async def on_message(message: Message) -> None:
-    global controllers
-    global loaded_image
+# Allow only one async routine to calculate list at a time
+list_task_control_lock = asyncio.Lock()
+
 
-    if message.author == client.user:
+# List command
+@client.slash_command()
```
anoadragon453, line 64: It would be cool to add a quick `description` to each command that's shown to users.

src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
+@application_checks.has_role(config.dj_role_name)
+async def list(
+    interaction: Interaction,
+    list_channel: Optional[TextChannel] = SlashOption(required=False),
```
anoadragon453, line 68: And descriptions for options.

src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
         return
 
-    # The message author must be a guild member, so that we can
-    # check if they have the appropriate role below
-    if not isinstance(message.author, Member):
+    if list_channel is None:
+        list_channel = interaction.channel
+
+    async with list_task_control_lock:
+        await interaction.response.defer(ephemeral=True)
```
anoadragon453, line 88: I think this makes the bot say "thinking..." right? If so, it'd be good to label that with a comment since it isn't immediately obvious from the code.

src/main.py
anoadragon453, line 68: I think we can limit `list_channel` to only suggest text channels in Discord's UI by using the `channel_types` argument to `SlashOption`: https://docs.nextcord.dev/en/stable/api.html#nextcord.SlashOption

`list_channel` needs to be of type `GuildChannel` though. So:

```py
list_channel: Optional[GuildChannel] = SlashOption(required=False, channel_types=[ChannelType.text]),
```

I don't know if having `Optional[GuildChannel]` instead of `GuildChannel` for the type will disable the checking code though.

src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
+        if bc is not None:
+            await interaction.delete_original_message()
+
+
+# Bust command
+@client.slash_command()
+@application_checks.has_role(config.dj_role_name)
+async def bust(
+    interaction: Interaction,
+    index: int = SlashOption(required=False, min_value=1, default=1),
```
anoadragon453, line 105:
```suggestion
    index: Optional[int] = SlashOption(required=False, min_value=1, default=1),
```

I think?

src/main.py
```diff
@@ -39,128 +56,170 @@ async def on_close() -> None:
...
+    await interaction.response.send_message("I didn't like that track anyways.")
+    bc.skip_song()
+
+
+# Stop command
+@client.slash_command()
+@application_checks.has_role(config.dj_role_name)
+async def stop(interaction: Interaction) -> None:
+    """Stop playback."""
+    bc = controllers.get(interaction.guild_id, None)
```
anoadragon453, line 150: Why not use `get_controller` here as well?

src/bust.py
```diff
@@ -327,125 +327,104 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+        list_message = await message_channel.send(embed=embed)
+        message_list.append(list_message)
+
+    # If message channel == target channel, or we ask for a mock,
+    # pin messages in reverse order and generate Google Form
+    pin_and_form = list_channel == message_channel
+    if pin_and_form:
+        for list_message in reversed(message_list):
+            await discord_utils.try_set_pin(list_message, True)
+        # Wrap form generation in try/catch so we don't block !list if it fails
```
anoadragon453, line 415:
```suggestion
        # Wrap form generation in try/catch so we don't block a list command if it fails
```

anoadragon453 (changes requested):
This is awesome! The usability is great, with command descriptions right in the app, and any errors that arise are only visible to the user that send the command. *It's like a whole new game.*

A grab bag of comments below, but overall looking pretty good!

src/bust.py
```diff
@@ -84,24 +82,24 @@ async def stop(self) -> None:
...
-            await message.channel.send(
-                "You need to be in an active voice or stage channel."
+            await interaction.response.send_message(
+                "You need to be in an active voice or stage channel.", ephemeral=True
             )
             return
 
         # Get a reference to the bot's Member object
-        bot_member = message.guild.get_member(self.client.user.id)
+        bot_member = interaction.guild.get_member(self.client.user.id)
```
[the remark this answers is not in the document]
Cephian: I'm a little confused, here `interaction.user` is the user who issued the command `/play` but `bot_member` is the user of the Busty bot herself. We use this variable to change Busty's nickname when songs are played etc.

src/main.py
Cephian: Setting the type hint automatically restricts to text channels automatically, [see docs](https://docs.nextcord.dev/en/stable/api.html#slash-options). I personally think it's easier to read as is now than what you suggested, but let me know if you think differently.

I'm also realizing now that the `required=False` argument is also deduced from the `Optional` type. Not sure if we should remove these, lmk what you think.

README.md
```diff
@@ -70,6 +70,7 @@ The complete list of environment variable configuration options is:
 1. `BUSTY_DJ_ROLE` - Name of role with permissions to run commands (default = bangermeister)
```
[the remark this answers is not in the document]
Cephian: I actually think what I said is correct, though it's very confusing. I think there is a strange distinction between command permissions and command visibility, and also a strange distinction between Discord-internal permissions (manage messages, kick members, etc) and role-based permissions.

The most clear thing I can find in writing is in the [Discord API Docs changelog](https://github.com/discord/discord-api-docs/blob/main/docs/Change_Log.md):

> You can enable or disable a command (guild or global) for a specific user or role in a guild. For now, users will still be able to see the commands, but won't be able to use them.

It was AFTER this update that the Integrations page was added, but I cannot find anything about bots being able to set their own command visibility.

When searching for things like "limit role slash command nextcord" I [got what I'm using here](https://stackoverflow.com/questions/72332580/how-do-you-make-a-specific-role-able-to-use-a-slash-command-in-nextcord) which is `application_checks.has_permissions`. This disallows a user without the proper role to RUN the command, but it still shows in the interface. In your link I cannot tell if "restricting access" is the same as "restricting visibility". 

Also, `default_member_permissions` [Permissions](https://docs.nextcord.dev/en/stable/api.html#nextcord.Permissions) object seems to only support native discord permissions (like manage_messages, kick_members, etc) and not custom role-based ones. I'm not sure if this is a limitation of Nextcord or of the Discord API itself.

This [huge GitHub thread](https://github.com/discord/discord-api-docs/issues/2315) is relevant and seems to only discuss solving the visibility issue by using the Integrations page.

src/main.py
Cephian: `@application_checks.has_role(config.dj_role_name)` automatically disables this command (and all the others) for DMs since it requires a role.

src/bust.py
```diff
@@ -327,125 +327,103 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
-    if message.channel_mentions:
-        target_channel = message.channel_mentions[0]
+    bc = BustController(client, channel_media_attachments, list_channel)
 
-    # Ensure target channel is text
-    if not isinstance(target_channel, TextChannel):
-        print(f"Cannot create controller for {type(target_channel)}")
-        await message.add_reaction(config.COMMAND_FAIL_EMOJI)
-        return None
+    # Title of !list embed
```
anoadragon453, line 346:
```suggestion
    # Title of /list embed
```

src/bust.py
```diff
@@ -327,125 +327,103 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+            )
+        else:
+            embed = Embed(
+                description=embed_description,
+                color=config.LIST_EMBED_COLOR,
             )
+        list_message = await interaction.channel.send(embed=embed)
+        message_list.append(list_message)
+
+    # If message channel == target channel, or we ask for a mock,
```
anoadragon453, line 408:
```suggestion
    # If message channel == target channel,
```

Since we removed this functionality for the time being.

src/main.py
```diff
@@ -19,17 +18,35 @@
 
 # Set up the Discord client. Connecting to Discord is done at
 # the bottom of this file.
-client = Client(intents=intents)
+if config.testing_server:
```
anoadragon453, line 21: Can we replace `testing_server` with `testing_guild`? As we use the "guild" terminology elsewhere.

anoadragon453 (changes requested):
Just small things now.

src/main.py
Cephian: I did my best to rename in the bizarre Discord tradition of calling them servers for users and guilds for developers, renaming all code references of "server" to guild (including the env variable) but calling them "server" in the user instructions in the README.

src/main.py
```diff
@@ -18,9 +18,9 @@
 
 # Set up the Discord client. Connecting to Discord is done at
 # the bottom of this file.
-if config.testing_server:
-    print(f"Using testing server {config.testing_server}")
-    ids = [int(config.testing_server)]
+if config.testing_guild:
+    print(f"Using testing guilde {config.testing_guild}")
```
anoadragon453, line 22:
```suggestion
    print(f"Using testing guild {config.testing_guild}")
```

anoadragon453 (approved):
LGTM!

PR #147, opened 2022-11-29
For-loop refactor

### LINKED ISSUE #123: Change control flow of playing next song

Cephian:
Closes #123.

Currently, the playback architecture treats the list of songs like a queue, popping each one off the front in a chain of coroutines which each call each other until the queue is empty.

This PR refactors it so that the list of songs is immutably traversed with a for-loop, and each playback event is a single coroutine call which returns when complete. This leads to a more modular and conceptually simpler control flow with less nonobvious "cleanup" corner cases. surrounding skipping and cancellation.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -41,7 +41,7 @@ class BustController:
         # Keep track of the coroutine used to play the next track, which is active while
         # waiting in the intermission between songs. This task should be interrupted if
         # stopping playback is requested.
-        self.play_next_task: Optional[asyncio.Task] = None
+        self.play_song_task: Optional[asyncio.Task] = None
 
         # The nickname of the bot. We need to store it as it will be
         # changed while songs are being played.
@@ -53,10 +53,8 @@ class BustController:
         self.current_channel_content: List[
             Tuple[Message, Attachment, str]
         ] = current_channel_content
-        # When play_next_task.cancel() is called, it is only actually cancelled on unsuspend
-        # so play_next_task.cancelled() may return False.
-        # We use this variable to keep track of /actual/ cancelled state
-        self.play_next_cancelled: bool = False
+        # Whether bust has been manually stopped
+        self.bust_stopped: bool = False
         # Client object
         self.client: Client = client
 
@@ -79,10 +77,13 @@ class BustController:
 
     async def stop(self) -> None:
         """Stop playing music."""
-        self.play_next_cancelled = True
-        if self.play_next_task:
-            self.play_next_task.cancel()
-        await self.finish(say_goodbye=False)
+        self.bust_stopped = True
+        if self.play_song_task:
+            self.play_song_task.cancel()
+
+    def skip_song(self) -> None:
+        if self.play_song_task:
+            self.play_song_task.cancel()
 
     async def play(self, message: Message, skip_count: int = 0) -> None:
         # Join active voice call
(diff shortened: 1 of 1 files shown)


src/bust.py
```diff
@@ -288,6 +271,9 @@ def ffmpeg_post_hook(e: BaseException = None):
         bot_member = self.current_channel.guild.get_member(self.client.user.id)
         await bot_member.edit(nick=new_nick)
 
+        # Wait for song to finish playing
+        await play_lock.acquire()
```
anoadragon453, line 272: What is the point of this lock?

Cephian: It waits for the song to finish playing before returning from the function, so that code like
```python
play_song(0)
play_song(1)
```
plays two songs in a row instead of trying to start playing two songs at almost the same time. It's necessary because `voice_client.play` is not async, it actually starts playback in a new thread.

src/bust.py
```diff
@@ -79,10 +77,13 @@ def finished(self) -> bool:
 
     async def stop(self) -> None:
         """Stop playing music."""
-        self.play_next_cancelled = True
-        if self.play_next_task:
-            self.play_next_task.cancel()
-        await self.finish(say_goodbye=False)
+        self.bust_stopped = True
+        if self.play_song_task:
```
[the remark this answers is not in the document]
Cephian: It looks like we could technically trigger this. When calling `bc.play()` there is a point after the voice client connects where Busty awaits to set herself as speaking in a stage channel (the line `await bot_member.edit(suppress=False)`), at which point control flow could switch to handling a stop command which would pass the `bc.is_active()` check in `main.py` and get to this line. Though admittedly I wasn't really thinking of this when writing, I did it so it's easier to check types and "just in case" since it's a cheap and readable future-proofing check.

src/bust.py
```diff
@@ -128,25 +130,22 @@ async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
...
+        # Play songs
+        for index in range(skip_count, len(self.current_channel_content)):
+            # wrap play_song() in a coroutine so it is cancellable
+            if self.bust_stopped:
+                break
+            self.play_song_task = asyncio.create_task(self.play_song(index))
+            try:
+                await self.play_song_task
+            except asyncio.CancelledError:
+                # Voice client playback must be manually stoped
```
anoadragon453, line 144:
```suggestion
                # Voice client playback must be manually stopped
```

src/bust.py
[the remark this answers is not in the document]
anoadragon453: Heh, yeah fair. It doesn't hurt to add either.

src/bust.py
```diff
@@ -128,25 +130,22 @@ async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
...
-        )
-        asyncio.run_coroutine_threadsafe(
-            self.play_next_task.get_coro(), self.client.loop
-        )
+        # Play songs
+        for index in range(skip_count, len(self.current_channel_content)):
+            # wrap play_song() in a coroutine so it is cancellable
+            if self.bust_stopped:
+                break
+            self.play_song_task = asyncio.create_task(self.play_song(index))
```
anoadragon453, line 140:
```suggestion
            if self.bust_stopped:
                break

            # wrap play_song() in a coroutine so it is cancellable
            self.play_song_task = asyncio.create_task(self.play_song(index))
```

src/bust.py
```diff
@@ -251,21 +234,18 @@ async def play_next_song(self, skip_count: int = 0) -> None:
...
-                self.play_next_coro()
+            # Unpin now playing message
+            if self.now_playing_msg:
+                await discord_utils.try_set_pin(self.now_playing_msg, False)
+                self.now_playing_msg = None
+            await play_lock.release()
 
         # Play song
+        play_lock = asyncio.Lock()
+        await play_lock.acquire()
```
anoadragon453, line 248: Why is this `await play_lock.acquire()` needed when we also do the same below?

anoadragon453 (approved):

src/bust.py
Cephian: Maybe I should write a little comment about the flow here because it might be a bit confusing with no context. The order of operations is

1. Acquire lock
2. Start playback
3. Try to acquire lock again, but get blocked because it's already acquired
4. Eventually the song ends, and it releases the lock
5. The second `acquire()` now gets unblocked, and the function returns

Without the first `acquire()`, that acquire before returning would succeed. In order for it to block until playback finishes, it needs to be acquired during the course of playback and then released when playback is done.

Cephian: Added clarifying comment
```python
# Acquire the lock during playback so that on release, play_song() returns
```

PR #150, opened 2022-12-14
Add /info command

### LINKED ISSUE #134: Create !stats command
Create a stats command intended to be run in private to tell admins various useful things about the bust. Some examples might be:

- Total bust time/total song time
- Total time remaining (when run during bust)
- Ending time (when run during bust)
- Top 3 people who have submitted the most content + those times
- How many submissions have album art/extra text

Cephian:
Closes #134. Must be used after `/list`. In case you were wondering, the @ mention DOES work even when the user is not in the channel or even on the server anymore.

image

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -349,7 +349,7 @@ class BustController:
                 f"*Total track length:* {song_utils.format_time(songs_len)}",
                 f"*Total bust length:* {song_utils.format_time(bust_len)}",
                 f"*Unique submitters:* {len(submitter_to_len)}",
-                f"*Longest submitter:* {longest_submitter.mention} : "
+                f"*Longest submitter:* {longest_submitter.mention} - "
                 + f"{song_utils.format_time(longest_submitter_time)}",
             ]
         )
--- a/src/bust.py
+++ b/src/bust.py
@@ -326,7 +326,7 @@ class BustController:
         )
         return form_url
 
-    async def send_stats(self, interaction: Interaction):
+    async def send_stats(self, interaction: Interaction) -> None:
         songs_len = int(self.total_song_len)
         num_songs = len(self.current_channel_content)
         bust_len = songs_len + config.seconds_between_songs * num_songs
--- a/src/bust.py
+++ b/src/bust.py
@@ -331,7 +331,7 @@ class BustController:
         num_songs = len(self.current_channel_content)
         bust_len = songs_len + config.seconds_between_songs * num_songs
 
-        # Compute map submitter --> total length of all submissions
+        # Compute map of submitter --> total length of all submissions
         submitter_to_len = defaultdict(lambda: 0.0)
 
         for submit_message, attachment, local_filepath in self.current_channel_content:
--- a/src/bust.py
+++ b/src/bust.py
@@ -1,5 +1,6 @@
 import asyncio
 import random
+from collections import defaultdict
 from typing import List, Optional, Tuple, Union
 
 from nextcord import (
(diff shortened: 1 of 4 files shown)


src/bust.py
```diff
@@ -325,6 +326,40 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
         )
         return form_url
 
+    async def send_stats(self, interaction: Interaction):
```
anoadragon453, line 329:
```suggestion
    async def send_stats(self, interaction: Interaction) -> None:
```

src/bust.py
```diff
@@ -325,6 +326,40 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+        songs_len = int(self.total_song_len)
+        num_songs = len(self.current_channel_content)
+        bust_len = songs_len + config.seconds_between_songs * num_songs
+
+        # Compute map submitter --> total length of all submissions
```
anoadragon453, line 334:
```suggestion
        # Compute map of submitter --> total length of all submissions
```

nit: it took me a second to parse in my head what "Compute map submitter" meant.

src/bust.py
```diff
@@ -325,6 +326,40 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+                f"*Number of tracks:* {num_songs}",
+                f"*Total track length:* {song_utils.format_time(songs_len)}",
+                f"*Total bust length:* {song_utils.format_time(bust_len)}",
+                f"*Unique submitters:* {len(submitter_to_len)}",
+                f"*Longest submitter:* {longest_submitter.mention} : "
+                + f"{song_utils.format_time(longest_submitter_time)}",
+            ]
+        )
+        embed = Embed(
+            title="Listed Statistics",
```
anoadragon453, line 357: What do you think of "Bust Statistics" instead?

src/bust.py
```diff
@@ -325,6 +326,40 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+        longest_submitter = max(submitter_to_len, key=submitter_to_len.get)
+        longest_submitter_time = int(submitter_to_len[longest_submitter])
+
+        embed_text = "\n".join(
+            [
+                f"*Number of tracks:* {num_songs}",
+                f"*Total track length:* {song_utils.format_time(songs_len)}",
+                f"*Total bust length:* {song_utils.format_time(bust_len)}",
+                f"*Unique submitters:* {len(submitter_to_len)}",
+                f"*Longest submitter:* {longest_submitter.mention} : "
```
anoadragon453, line 352:
```suggestion
                f"*Longest submitter:* {longest_submitter.mention} - "
```

nit: since we already have a colon on this line.

src/main.py
```diff
@@ -217,6 +217,22 @@ async def image_view(interaction: Interaction) -> None:
...
+@application_checks.has_role(config.dj_role_name)
+async def info(interaction: Interaction) -> None:
+    """Get info about currently listed songs."""
+    bc = controllers.get(interaction.guild_id)
+
+    if bc is None:
+        await interaction.response.send_message(
+            "You need to use /list first", ephemeral=True
+        )
+        return
```
anoadragon453, line 231: I wonder if we should automatically create a bust here / overwrite the current one with the latest information if `/list` has already been called. Otherwise we'll always need to `/list` manually first, or the info could be about an old, outdated `/list` invocation.

anoadragon453 (commented):
This looks pretty good on the whole. I think the following from the linked issue could be useful as well for planning post-bust things:

* Ending time (when run during bust)

But could also be done in a separate PR.

src/main.py
Cephian: If we give /info the same interface for passing in which channel to list I think it's a little confusing, in particular running `/info` with no arguments would depend on whether `/list` has been run for which channel it generates stats for. If something has been listed, then that channel. Otherwise, the one the message was sent in.

Maybe if you run with no arguments you MUST have already listed, but if you run with a channel argument you re-list that channel and then show info. I think it's a good idea, but I also think it should be a separate PR since it's basically just adding a feature on top of this one. If this sounds good to you we can make an issue and I'll get to it soonish

src/bust.py
Cephian: Sounds much better, I had a hard time coming up with what should go here lol

Cephian:
> * Ending time (when run during bust)
I love this idea, and yes separate PR as it will require changes in more parts of the code. I''ll make an issue

src/main.py
anoadragon453: Let's try it out in its current form and see whether it'd be a useful addition.

anoadragon453 (approved):

PR #153, opened 2023-01-03
Add a command which lets admins send announcements as the bot

anoadragon453:
This commit adds an `/announce` command which lets admins send an embed as the bot into any channel with a given title and body.

image
image
image

You can optionally specify a channel to send the announcement in. If a channel other than the current channel is specified, then the bot will respond with an ephemeral message confirming success of the action. If the current channel is specified, then no ephemeral message will be sent (as the admin will immediately see the announcement anyways).

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -258,24 +258,18 @@ async def announce(
     # Disallow sending announcements from one guild into another.
     if channel.guild.id != interaction.guild_id:
         interaction.response.send_message(
-            "Sending announcements to a guild outside of this channel is not allowed."
+            "Sending announcements to a guild outside of this channel is not allowed.",
+            ephemeral=True,
         )
         return
 
-    # If this command was invoked in the same channel the announcement is intended to be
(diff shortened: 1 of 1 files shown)


Cephian (approved):
Going to approve and merge, but leaving a summary of changes I made here
image

1) I made the "you can't send messages to another guild" message ephemeral, as "you can't do that" messages are ephemeral elsewhere.

2) I changed the behavior so that it looks the same to other people whether you included the optional command argument or not. I don't think announcements should look different for the user depending on whether the channel argument was used, and I think it breaks immersion to reply to the command-issuer since these announcements are supposed to be from Busty herself.

anoadragon453:
> I think it breaks immersion to reply to the command-issuer since these announcements are supposed to be from Busty herself.

Aha, I didn't realise there would be a little reply-to indicator when sending in the same channel. Good catch, we indeed do not want that to appear.

PR #154, opened 2023-01-03
Handle song length errors without crashing

Cephian:
Busty crashes on `/info` if `get_song_length()` returns None (which it can). Instead, we should treat that song's length as zero and report that statistics may be inaccurate due to errors.

image

To deal with the unlikely but possible case that ALL songs have errors, we introduce a "phantom submitter" which is the bot member herself, with a total submission length of zero.

image

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -316,11 +316,18 @@ class BustController:
 
         # Compute map of submitter --> total length of all submissions
         submitter_to_len = defaultdict(lambda: 0.0)
+        # Default entry in case no valid songs w/ length
+        bot_member = interaction.guild.get_member(self.client.user.id)
+        submitter_to_len[bot_member] = 0.0
 
+        errors = False
         for submit_message, attachment, local_filepath in self.bust_content:
             song_len = song_utils.get_song_length(local_filepath)
-            submitter = submit_message.author
-            submitter_to_len[submitter] += song_len
+            if song_len:
+                submitter = submit_message.author
+                submitter_to_len[submitter] += song_len
+            else:
+                errors = True
 
         # User with longest total submission length
         longest_submitter = max(submitter_to_len, key=submitter_to_len.get)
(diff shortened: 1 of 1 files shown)


src/bust.py
```diff
@@ -331,11 +338,15 @@ async def send_stats(self, interaction: Interaction) -> None:
                 f"*Number of tracks:* {num_songs}",
                 f"*Total track length:* {song_utils.format_time(songs_len)}",
                 f"*Total bust length:* {song_utils.format_time(bust_len)}",
-                f"*Unique submitters:* {len(submitter_to_len)}",
+                f"*Unique submitters:* {len(submitter_to_len)-1}",
```
anoadragon453, line 341:
```suggestion
                # We account for the bot being included in the `submitter_to_len` dict
                # by subtracting 1 from the total count.
                f"*Unique submitters:* {len(submitter_to_len)-1}",
```

anoadragon453 (approved):
This looks good to me. An alternative to including the bot member in the `submitter_to_len` dict is to always include submitters in the `dict`, but just give them an entry of 0 if we fail to read the song length.

Cephian:
> This looks good to me. An alternative to including the bot member in the `submitter_to_len` dict is to always include submitters in the `dict`, but just give them an entry of 0 if we fail to read the song length.

I think I prefer it this since then there's no need to add another loop pass to check which ones are empty. And considering this is a case that will /never/ come up in practice, less obscure/time consuming logic seems better for being technically correct.

anoadragon453:
Hmm, I've also just realised that the "Unique submitters" count will be wrong if that submitter's song couldn't have the length read.

Cephian:
> Hmm, I've also just realised that the "Unique submitters" count will be wrong if that submitter's song couldn't have the length read.

Oh yeah, I was thinking that this was fine since I put the "some stats may be inaccurate" but it's probably better to just make sure an entry for that user is created even if they have zero length. that way we won't even need the Busty submitter since iirc you can't list with no submissions

anoadragon453:
I think that's the same as what I [said above](https://github.com/anoadragon453/busty/pull/154#pullrequestreview-1241376211) :P

Cephian:
> I think that's the same as what I [said above](https://github.com/anoadragon453/busty/pull/154#pullrequestreview-1241376211) :P

Yeah I interpreted it incorrectly haha, I'll patch this to that version at some point

PR #155, opened 2023-01-12
Better internal error handling for bad song len

Cephian:
As suggested by @anoadragon453 in https://github.com/anoadragon453/busty/pull/154#pullrequestreview-1241376211. Cleaner way to handle the case when no songs with readable length have been submitted.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -341,7 +341,7 @@ class BustController:
         )
         if errors:
             embed_text += (
-                "\n\n**There were some errors. Statistics may not be accurate.**"
+                "\n\n**There were some errors. Statistics may be inaccurate.**"
             )
         embed = Embed(
             title="Listed Statistics",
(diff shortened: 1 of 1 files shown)


src/bust.py
```diff
@@ -316,18 +316,14 @@ async def send_stats(self, interaction: Interaction) -> None:
...
         errors = False
         for submit_message, attachment, local_filepath in self.bust_content:
             song_len = song_utils.get_song_length(local_filepath)
-            if song_len:
-                submitter = submit_message.author
-                submitter_to_len[submitter] += song_len
-            else:
+            if song_len is None:
                 errors = True
+                song_len = 0
```
anoadragon453, line 325: pedantic, but just to keep the type of each value in this dict the same:

```suggestion
                song_len = 0.0
```

anoadragon453, line 325: Also, could you add a comment for our future selves explaining why we still add an entry to `submitter_to_len` even if we can't deduce the length?

anoadragon453 (changes requested):

anoadragon453 (approved):

PR #156, opened 2023-01-15
Store persistent bot state in a json file with a configurable path

anoadragon453:
This PR adds the ability to store arbitrary persistent state (survives across bot restarts), backed by a JSON store on disk. The store's location can be configured via the `BUSTY_BOT_STATE_FILE`, and defaults to `bot_state.json`.

State can be stored via `persistent_state.set_state(path, value)` and retrieved via `value = persistent_state.get_state(path)`. You can also delete state via `persistent_state.delete_state(path)`. Deleting state at a path will also clear any parent paths that have no state left in them after the delete, in order to keep the store neat and tidy.

The plan is to replace `BUSTY_IMAGE_STATE_FILE` and `PersistentString` with this persistent store implementation.

### DIFF
--- a/src/persistent_state.py
+++ b/src/persistent_state.py
@@ -1,0 +1,168 @@
+import copy
+import json
+from typing import Iterable
+
+from config import JSON_DATA_TYPE, bot_state_file
+
+# Global, persistent state of the bot. Not to be accessed directly. Instead, use the
+# getter and setter methods below.
+_bot_state = {}
+
+
+def load_state_from_disk() -> None:
+    """Read the bot state from disk and store it in memory."""
+    global _bot_state
+
+    with open(bot_state_file) as f:
+        bot_state_str = f.read()
+
+        if bot_state_str:
+            # Load the JSON representation of the bot's state and convert it to a Python dict so
+            # it can be easily manipulated.
+            _bot_state = json.loads(bot_state_str)
+
+
+def set_state(path: Iterable[str], value: JSON_DATA_TYPE) -> None:
+    """Modifies the bot's state in memory, then dumps the modified state to disk.
+
+    This function should not be made async! We do not want to yield from it before we update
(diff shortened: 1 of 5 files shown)


src/persistent_state.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # `current_path` is now set to the dictionary where we're like to place our `key` and our `value`.
+        # As `current_path` is just a reference to a dictionary inside `bot_state`, this modifies
+        # `bot_state` as well.
+        current_path[key] = value
+    else:
+        # If no path was provided, just override all bot state with the given value.
+        # The given value must be a dict.
+        if not isinstance(value, dict):
+            raise Exception("Attempted to override all bot state with a non-dict type")
```
Cephian, line 70:
```suggestion
            raise Exception("Attempted to override entire bot state with a non-dict type")
```
Just because I like the sound of it better 8^)

Cephian (approved):

src/persistent_state.py
```diff
@@ -0,0 +1,168 @@
...
+        return False
+
+    if field_to_delete not in state_at_path:
+        # This field does not exist at this path.
+        return False
+
+    del state_at_path[field_to_delete]
+
+    # Store the updated state to disk
+    set_state(path, state_at_path)
```
Cephian, line 161: Actually, shouldn't we save to disk after deleting the path in the following lines?

Cephian (changes requested):

src/persistent_state.py
anoadragon453: This is what I initially tried, but note that `state_at_path` here is a copy of the bot state dict, not the dict itself. If we don't update the actual bot state here, then the call to `delete_state` just below will end up calling `get_state` again, yielding a dictionary without any deleted state.

I can add a comment explaining this as it is subtle.

Cephian (approved):

PR #157, opened 2023-01-15
Store Google form image URLs via bot state

anoadragon453:
Note: This builds on top of (and includes the commits of) https://github.com/anoadragon453/busty/pull/156. To only review the commits unique to this PR, start from c56f26dd53b3d6621bd8d42d61acc40b100c4430.

This PR replaces `PersistentString` with the bot state feature from https://github.com/anoadragon453/busty/pull/156. The image URL for a guild's google form is now stored at `guilds` -> `<guild ID>` -> `form_image_url`. The guild ID is taken from the interaction.

I initially considered putting the form helper methods in `main.py`, but figured it'd be best to keep that file limited to only slash commands and other global setup. I also considered adding another `*_utils.py` file, but then felt it best to just keep helper methods related to persistent state in `persistent_state.py`.


src/main.py
```diff
@@ -177,44 +171,60 @@ async def image(interaction: Interaction) -> None:
...
     await interaction.response.send_message(
-        f"\N{WHITE HEAVY CHECK MARK} Image set to: {loaded_image.get()}"
+        f"\N{WHITE HEAVY CHECK MARK} Image set to: {loaded_image_url}"
     )
 
 
 @image.subcommand(name="url")
 @application_checks.has_role(config.dj_role_name)
-async def image_url(interaction: Interaction, image_url: str) -> None:
+async def image_by_url(interaction: Interaction, image_url: str) -> None:
```
anoadragon453, line 189: I renamed this function as technically `image_url` in the function's arguments overwrites the symbol `image_url`. So it'd technically be impossible to access a reference to the function before this change.

(Mostly I just did it because my IDE was yelling at me.)

src/persistent_state.py
```diff
@@ -166,3 +168,56 @@ def delete_state(path: Iterable[str]) -> bool:
...
+    Returns:
+        True if the image url saved correctly, False if not.
+    """
+    try:
+        set_state(["guilds", str(interaction.guild_id), "form_image_url"], image_url)
+    except Exception as e:
+        print("Unable to set form image:", e)
+
+        await interaction.response.send_message(
+            f"Failed to upload image ({type(e)}). See the logs for more details",
```
Cephian, line 192: If the first sentence end with a period, the second one probably should also. This also applies to the other two instances.

I'm actually realizing now we're sort of inconsistent across the codebase about whether Busty's sentences end with periods, we should probably fix that

src/main.py
```diff
@@ -177,44 +171,60 @@ async def image(interaction: Interaction) -> None:
...
     """Upload a Google Forms image as attachment."""
-    global loaded_image
     # TODO: Some basic validity filtering
-    loaded_image.set(image_file.url)
+    # Persist the image URL
+    if not await persistent_state.save_form_image_url(interaction, image_file.url):
+        return
+
+    # Now load it again to ensure that it has saved correctly
+    loaded_image_url = persistent_state.get_form_image_url(interaction)
```
Cephian, line 180: In what sense does this "ensure that it has saved correctly" ? It seems more like you're loading it just to have a more readable variable name, and calling `get_form_image_url()` instead of using `image_file.url` for better logical separation of parts.

I guess maybe if the command throws an exception then something went wrong, but that feels like something that should happen in the save function

src/persistent_state.py
anoadragon453: > If the first sentence end with a period, the second one probably should also. This also applies to the other two instances.

Generally I don't agree with this. I think it's fine to have error messages without periods at the very end, but still with periods in between sentences to separate them.

With respect to the busty codebase I don't feel too strongly about whether we end our error messages with periods or not as long as we're consistent. I agree that since these error messages end up being displayed as chat messages, ending them with punctuation comes across nicely.

I've added the missing periods in 4419dc9.

src/main.py
anoadragon453: Well, this is what we had before already with the call to `loaded_image.get()`.

It's essentially an automated call to `/image view`, which we currently always do after setting the image - just in case. But I agree that technically it shouldn't be necessary, as `save_form_image_url` should fail if the image wasn't set correctly.

We should really just trust that `/image` commands work if they don't fail and avoid the extra CPU cycles.

anoadragon453: Removed the loading in c3cb3ec.

Cephian (approved):

PR #160, opened 2023-01-15
Use `discord_emoji` library directly to stay up-to-date with new emoji

anoadragon453:
New unicode emoji are occasionally released, and Discord updates the set of emoji they support accordingly. The [discord-emoji library](https://github.com/sevenc-nanashi/discord-emoji) (which we previously just copy-pasted the list of emoji from) continues to update itself to support new emoji. If we wanted to do the same, we would have to continue copy-pasting and removing our blacklisted emoji from the list. That is both time-consuming, and prone to errors as we didn't have a list written down of the emoji that we wanted to remove.

I reproduced the list of blacklisted emoji from these commits (https://github.com/anoadragon453/busty/pull/10, https://github.com/anoadragon453/busty/pull/24).

**Note:** This adds a new dependency, so we will need to `pip install` in production after this has merged. Though as the library is imported at startup, not doing so produce an obvious error.


src/bust.py
```diff
@@ -197,7 +197,7 @@ async def play_song(self, index: int) -> None:
         submit_message, attachment, local_filepath = self.bust_content[index]
 
         # Associate a random emoji with this song
-        random_emoji = random.choice(config.emoji_list).encode("Latin1").decode()
+        random_emoji = random.choice(config.emoji_list)
```
anoadragon453, line 200: Doing this encoding and decoding dance is no longer necessary after the library [updated to encode unicode properly](https://github.com/sevenc-nanashi/discord-emoji/commit/7699082118ba22d4e007d1f48be1f47c48e32de4).

Cephian (approved):

PR #164, opened 2023-01-18
Add some missing periods where they should be

Cephian:
This likely has merge conflicts with some currently waiting pull requests, but it's simple enough that with approval I'll just merge in and push myself after those are pushed.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -161,7 +161,7 @@ async def stop(interaction: Interaction) -> None:
         await interaction.response.send_message("Nothing is playing.", ephemeral=True)
         return
 
-    await interaction.response.send_message("Alright I'll shut up")
+    await interaction.response.send_message("Alright I'll shut up.")
     await bc.stop()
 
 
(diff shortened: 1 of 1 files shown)


jcogan97 (approved):

src/main.py
```diff
@@ -161,7 +161,7 @@ async def stop(interaction: Interaction) -> None:
         await interaction.response.send_message("Nothing is playing.", ephemeral=True)
```
anoadragon453, line 161: There's a couple missing periods still at https://github.com/anoadragon453/busty/blob/8ca62a58a34faa6ec996e6e6515b525f92631bf4/src/main.py#L196

https://github.com/anoadragon453/busty/blob/8ca62a58a34faa6ec996e6e6515b525f92631bf4/src/main.py#L214

(and you may want to remove the colons as well as you did below.)

anoadragon453 (approved):
lgtm otherwise!

PR #166, opened 2023-01-31
Don't crash when starting with no bot state file

Cephian:
Right now if there is no bot_state.json, Busty's won't start. This is bad because the initial run will have no bot_state.json. This PR changes it so that if reading the bot_state.json file fails, a warning is printed and the bot continues with empty state.

### DIFF
--- a/src/persistent_state.py
+++ b/src/persistent_state.py
@@ -18,7 +18,7 @@ def load_state_from_disk() -> None:
         with open(bot_state_file) as f:
             bot_state_str = f.read()
     except (FileNotFoundError, IOError):
-        print(f'Could not read state from {bot_state_file}. Continuing...')
+        print(f"Could not read state from {bot_state_file}. Continuing...")
 
     if bot_state_str:
         # Load the JSON representation of the bot's state and convert it to a Python dict so
--- a/src/persistent_state.py
+++ b/src/persistent_state.py
@@ -17,8 +17,12 @@ def load_state_from_disk() -> None:
     try:
         with open(bot_state_file) as f:
             bot_state_str = f.read()
-    except (FileNotFoundError, IOError):
-        print(f"Could not read state from {bot_state_file}. Continuing...")
+    except FileNotFoundError:
+        # Expected on first run or after setting a new custom bot state filepath.
+        pass
+    except IOError:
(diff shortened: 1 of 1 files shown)


src/persistent_state.py
```diff
@@ -13,13 +13,17 @@ def load_state_from_disk() -> None:
...
-        if bot_state_str:
-            # Load the JSON representation of the bot's state and convert it to a Python dict so
-            # it can be easily manipulated.
-            _bot_state = json.loads(bot_state_str)
+    bot_state_str = None
+    try:
+        with open(bot_state_file) as f:
+            bot_state_str = f.read()
+    except (FileNotFoundError, IOError):
+        print(f"Could not read state from {bot_state_file}. Continuing...")
```
anoadragon453, line 21: I think we shouldn't print a user-visible error in the case of the `bot_state.json` file not existing, as that will be the state that everyone experiences on first run (or upon setting a custom location).

I think we shouldn't continue on if we fail to read the file as well, as the bot will pretty quickly overwrite the contents of the file with mostly-empty data. I think in that case we should fail loudly and print the exception.

```suggestion
    except FileNotFoundError:
        # Expected on first run or after setting a new custom bot state filepath.
        pass
    except IOError:
        print(f"Could not read state from {bot_state_file}")
        raise
```

src/persistent_state.py
```diff
@@ -13,13 +13,17 @@ def load_state_from_disk() -> None:
...
     global _bot_state
 
-    with open(bot_state_file) as f:
-        bot_state_str = f.read()
-
-        if bot_state_str:
-            # Load the JSON representation of the bot's state and convert it to a Python dict so
-            # it can be easily manipulated.
-            _bot_state = json.loads(bot_state_str)
+    bot_state_str = None
```
anoadragon453, line 16: If the other suggestion is taken, then `bot_state_str` will always be set.

```suggestion
```

anoadragon453 (changes requested):

src/persistent_state.py
Cephian: I was thinking of it as more of a log message than an error, but yeah it's probably not information which is terribly important

anoadragon453 (approved):

PR #167, opened 2023-02-03
Add a 'preview' command that lets you preview your submission before a bust starts

### LINKED ISSUE #144: Add a 'preview' command that lets you preview your submission before a bust starts
Assuming #143 is implemented and we have support for slash commands, it would be nice to have a `/preview` command that everyone could use. One would use `/preview` to submit an audio file and submission text to the busty bot, and the bot would respond with a message that only you can read that shows what your submission will look like when it plays live during a bust.

Most importantly, users can check that their album art will be read correctly and that it looks as they expect. When they are happy with their submission, they submit their track by uploading it to busty's mailbox just as they have always done.

Proposed command syntax:

```
/preview audio:<audio file> <text>
```

an example:

```
/preview audio:<audio file> This is a song I made with my main man @qbert. What a crazy fellow he is. Vocal help from @J. Cole.
```

Discord should pop up a prompt to select a file from your OS for the audio file. I'm hoping it can do so automatically and users won't need to type/select `audio` manually.

Additionally, once #145 is implemented, it'd be nice to include those checks in the output of this command as well.

jcogan97:
Closes #144.

`/preview` will return a private embedded object to the user showing all the details of their uploaded file.

Using the Command:
CleanShot 2023-02-03 at 08 39 21@2x


src/main.py
```diff
@@ -232,8 +235,35 @@ async def info(interaction: Interaction) -> None:
...
+    submit_message: str = SlashOption(required=False, description="The text for submission message.")
+) -> None:
+    """Send a message to user with embedded preview of uploaded song's metadata."""
+    user = interaction.user
+    attachment_filepath = filepath_builder(interaction.id, uploaded_file)
+    await uploaded_file.save(fp=attachment_filepath)
+    
+    # TODO: Catch file types that are not media files containing .wav, .mp3, etc.
+    # For now, we'll just have if file is None.
+    if uploaded_file.content_type is None:
```
Cephian, line 261: There's no TODO here, we should make preview parsing follow the same logic as submission parsing on launch. in discord_utils.py we can see that the exact condition for whether we try to parse a file is 

```
            if attachment.content_type is None or (
                not attachment.content_type.startswith("audio")
                and not attachment.content_type.startswith("video")
            ):
```

any other invalidities should be caught at a later stage. This if-statement is complex enough that i'd probably extract it out into a function

src/main.py
```diff
@@ -295,4 +325,4 @@ async def on_application_command_error(
 else:
     print(
         "Please pass in a Discord bot token via the BUSTY_DISCORD_TOKEN environment variable."
-    )
+    )
```
Cephian, line 336: nitpick, stay consistent with formatting stuff like this

src/bust.py
```diff
@@ -82,7 +84,7 @@ def skip_song(self) -> None:
         """Skip current track."""
         if self.play_song_task:
             self.play_song_task.cancel()
-
+
```
Cephian, line 87: nitpick, stay consistent with formatting stuff like this

src/song_utils.py
```diff
@@ -8,11 +9,55 @@
...
 from mutagen.ogg import OggFileType
 from mutagen.wave import WAVE
-from nextcord import File
+from nextcord import File, Embed, Message, Attachment, User, Member
+from nextcord.utils import escape_markdown
 from PIL import Image, UnidentifiedImageError
 
 import config
 
+def embed_song(submit_message: Message | str, attachment_filepath: str, attachment: Attachment, user: User | Member, random_emoji=None) -> Embed:
```
Cephian, line 19: I think it would be cleaner to call "random_emoji" just "emoji" and make it nonoptional, forcing the caller to provide the random one if they are missing it

src/song_utils.py
```diff
@@ -8,11 +9,55 @@
...
+    
+    if isinstance(submit_message, Message):
+        message: str = submit_message.content
+    else:
+        message: str = submit_message
+        
+    emoji = random_emoji if random_emoji != None else random.choice(config.emoji_list)
+    
+    # Build and send "Now Playing" embed
+    if isinstance(submit_message, Message) == False:
```
Cephian, line 29: This function duplicates a lot of code, it's almost two copies of the same piece of code with a few if statements. You should try to find some better ways to combine similar lines.

I also think you should not change things in the format like Playing -> Previewing and removing the jump. I realize that a jump link doesn't make sense, but I think it should still be there and point to some sensible default URL. Things should be as visually identical as possible

Cephian (changes requested):
Not everything I have to say, just some general stuff to start with

src/main.py
```diff
@@ -234,6 +238,50 @@ async def info(interaction: Interaction) -> None:
...
+    attachment_filepath = filepath_builder(interaction.id, uploaded_file)
+    random_emoji = random.choice(config.emoji_list)
+    await uploaded_file.save(fp=attachment_filepath)
+
+    if is_valid_media(uploaded_file.content_type) is False:
+        await interaction.response.send_message(
+            "Sorry, looks like you didn't send the correct media type. \nTry that one again, kid.",
+            ephemeral=True,
+        )
+    else:
```
Cephian, line 263: To avoid unnecessary indent I'd just make it return in the if statement and have no else

src/song_utils.py
```diff
@@ -8,12 +8,52 @@
...
 
+def embed_song(
+    submit_message: Message | str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User | Member,
+    random_emoji,
+    jump_url=None,
+) -> Embed:
+    if isinstance(submit_message, Message) is False:
```
Cephian, line 26: Instead of depending on the type of submit message and having this weird jump_url optional [parameter,](url) I think the `jump_url` param should be mandatory, and `submit_message` should be renamed to `message_content` and should always be a string. Then when embed_song is called the caller must manually pass in the correct params

src/song_utils.py
```diff
@@ -8,12 +8,52 @@
...
+        escape_markdown(song_format(attachment_filepath, attachment.filename)),
+        attachment.url,
+        url,
+    )
+    embed = Embed(
+        title=embed_title, description=embed_content, color=config.PLAY_EMBED_COLOR
+    )
+
+    if message:
+        if len(message) > config.EMBED_FIELD_VALUE_LIMIT:
```
Cephian, line 48: I'm not sure why you unwrapped what was there before with a single if into an if-else, it causes code duplication in the last line. You should make it just an if and share the last line again

src/discord_utils.py
```diff
@@ -31,6 +31,29 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+        "{}.{}{}".format(
+            message_id,
+            attachment.id,
+            os.path.splitext(attachment.filename)[1],
+        ),
+    )
+    return filepath
+
+
+def is_valid_media(attachment_content_type: str) -> bool:
```
Cephian, line 47: Instead of `if (b) return false else return true,` just do `return not b`. In this case you'll should propagate the not over the rest of the condition, don't just surround it in parens and put not

Cephian (changes requested):
some more changes, still probably not complete

src/song_utils.py
jcogan97: I changed this due to the edge case with going between the Message object and a string for the `message_content`.

With your suggested changes, this can go back to how it was.

src/discord_utils.py
```diff
@@ -45,13 +45,10 @@ def filepath_builder(message_id: int, attachment: Attachment) -> str:
 
 
 def is_valid_media(attachment_content_type: str) -> bool:
-    if attachment_content_type is None or (
-        not attachment_content_type.startswith("audio")
+    return attachment_content_type is None or (
```
Cephian, line 48: This is not logically equivalent to what you changed it from

src/song_utils.py
```diff
@@ -16,40 +16,32 @@
 
 
 def embed_song(
-    submit_message: Message | str,
+    message_content: Message | str,
```
Cephian, line 19: should have type just str now

src/song_utils.py
```diff
@@ -16,40 +16,32 @@
...
     attachment_filepath: str,
     attachment: Attachment,
     user: User | Member,
     random_emoji,
```
Cephian, line 23: add types for random emoji and jump url

src/song_utils.py
```diff
@@ -16,40 +16,32 @@
...
         title=embed_title, description=embed_content, color=config.PLAY_EMBED_COLOR
     )
 
-    if message:
-        if len(message) > config.EMBED_FIELD_VALUE_LIMIT:
-            more_info = message[: config.EMBED_FIELD_VALUE_LIMIT - 1] + "…"
-            embed.add_field(name="More Info", value=more_info, inline=False)
-        else:
-            embed.add_field(name="More Info", value=message, inline=False)
+    if message_content:
```
Cephian, line 40: imo the more_info variable is redundant and you can edit message_content directly here

src/main.py
```diff
@@ -1,12 +1,16 @@
 import asyncio
+import random
+from os import remove
```
Cephian, line 3: from os import remove is a little confusing, just import os and do os.remove

src/main.py
```diff
@@ -1,12 +1,16 @@
...
 from typing import Dict, Optional
 
 from nextcord import Attachment, Embed, Intents, Interaction, SlashOption, TextChannel
 from nextcord.ext import application_checks, commands
 
 import config
 import persistent_state
+import song_utils
 from bust import BustController, create_controller
+from discord_utils import filepath_builder, is_valid_media
```
Cephian, line 13: I would say for these also don't import the function, keep discord_utils when you call them. It can be confusing to find where they come from otherwise

Cephian (changes requested):

src/song_utils.py
```diff
@@ -8,12 +8,43 @@
...
 from PIL import Image, UnidentifiedImageError
 
 import config
 
 
+def embed_song(
+    message_content: str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User | Member,
```
Cephian, line 22: Member is a subclass of user so there is no need for a union type. Choose the most general type applicable

Cephian (dismissed):

src/song_utils.py
```diff
@@ -8,12 +8,43 @@
...
 
 import config
 
 
+def embed_song(
+    message_content: str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User,
+    random_emoji: str,
```
Cephian, line 23: change this to just "emoji", it's not random within this function as its passed in

src/main.py
```diff
@@ -237,6 +241,50 @@ async def info(interaction: Interaction) -> None:
     await bc.send_stats(interaction)
 
 
+# Preview command
+@client.slash_command(dm_permission=False)
+async def preview(
+    interaction: Interaction,
+    uploaded_file: Attachment = SlashOption(description="The file to submit."),
+    submit_message: str = SlashOption(
```
Cephian, line 249: If required is false the type of this should be Optional[str]

src/config.py
```diff
@@ -27,6 +27,8 @@
 VOLUME_MULTIPLIER = 0.5
 # The maximum number of songs to download concurrently
 MAXIMUM_CONCURRENT_DOWNLOADS = 8
+# Link to Rick Astley's Never Gonna Give You Up
+YOUTUBE_URL = "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
```
Cephian, line 31: Don't describe the value of the variable, describe the semantics of what it holds. also YOUTUBE_URL is probably not the best name. The comment and variable name should describe what this variable does, not what the string value is. e.g. the var might be better named PREVIEW_JUMP_URL and the comment saying something about it being the default jump url for the preview embed

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def filepath_builder(message_id: int, attachment: Attachment) -> str:
```
Cephian, line 34: Maybe better to rename this slightly more semantically to "attachment_local_filepath" or something to better reflect that it constructs a unique local filepath from an attachment. Also maybe add a comment/docstring explaining this

Cephian (dismissed):

src/song_utils.py
```diff
@@ -8,12 +8,43 @@
...
+        attachment.url,
+        jump_url,
+    )
+    embed = Embed(
+        title=embed_title, description=embed_content, color=config.PLAY_EMBED_COLOR
+    )
+
+    if message_content:
+        if len(message_content) > config.EMBED_FIELD_VALUE_LIMIT:
+            message_content[: config.EMBED_FIELD_VALUE_LIMIT - 1] + "…"
```
Cephian, line 42: also this needs to have a `message_content = ` in front of it, right now it does nothing

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def attachment_local_filepath(message_id: int, attachment: Attachment) -> str:
+    # Computed local filepath
```
Cephian, line 35: this comment doesn't really explain anything, can you add a short comment explaining why it's needed and what it does (assign unique local filename to remote discord attachment etc)

src/config.py
```diff
@@ -27,6 +27,8 @@
 VOLUME_MULTIPLIER = 0.5
 # The maximum number of songs to download concurrently
 MAXIMUM_CONCURRENT_DOWNLOADS = 8
+# Holds a default URL value jump url for jump_url parameter in embed_song.
+DEFAULT_JUMP_URL = "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
```
Cephian, line 31: calling it "default" isn't really accurate, this is the jump url used for previews. it's not really default, if anything it's the non-default since the default is the jump url to the real message

Cephian (changes requested):

Cephian (changes requested):
Something I probably should have mentioned earlier is that you should add this command to the README

Cephian (changes requested):
Did you read the main text of my last review about the readme? make sure to not just read per-file comments

README.md
```diff
@@ -103,25 +104,26 @@ It should connect to Discord and display the currently logged-in application nam
...
-* Users comment on songs while they play.
-* An admin can run `!skip` at any time to skip the current song, or `!stop` to manually stop the show.
-* Once the last song has played, the bot will post a concluding message and leave the call.
+- Users submit songs into a channel.
+- All users join a voice channel or stage.
+- An admin runs `!list` to list all submitted songs and the order they will be played in.
+- An admin runs `!bust` to start the show. The bot will join the channel and begin playing songs in the order they were submitted.
+- Users comment on songs while they play.
+- An admin can run `!skip` at any time to skip the current song, or `!stop` to manually stop the show.
+- Once the last song has played, the bot will post a concluding message and leave the call.
```
jcogan97, line 113: VSCode keeps converting these on save

Cephian (dismissed):
getting close :^) I'm fine with all the random formatting changes your vscode seems to have done

README.md
anoadragon453: heh, you can turn this off by disabling the "Format on save" option in VSCode's preferences btw.

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+        "{}.{}{}".format(
+            message_id,
+            attachment.id,
+            os.path.splitext(attachment.filename)[1],
+        ),
+    )
+    return filepath
+
+
+def is_valid_media(attachment_content_type: str) -> bool:
```
anoadragon453, line 47: Could you include a docstring to explain what this function does (and what is considered "valid media")?

Something along the lines of:

```suggestion
def is_valid_media(attachment_content_type: str) -> bool:
"""
Returns whether an attachment's content type is considered "media".

Valid media types are those that start with "audio" or "video".

Args:
    attachment_content_type: A content type (aka MIME type).
Returns:
    True if the content type is considered to be media, otherwise False.
```

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+    return attachment_content_type is not None and (
```
anoadragon453, line 48: According to the type of `attachment_content_type` in the function's definition, it cannot be `None`. If it is possible that it could be `None`, its type should be `Optional[str]` instead.

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+        (attachment_content_type.startswith("audio"))
+        or (attachment_content_type.startswith("video"))
```
anoadragon453, line 50: These extra parentheses aren't necessary as they only surround a single entity.

```suggestion
        attachment_content_type.startswith("audio")
        or attachment_content_type.startswith("video")
```

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def attachment_local_filepath(message_id: int, attachment: Attachment) -> str:
+    """Build local filepath with provided Message ID & Attachment object for saving Attachments locally"""
```
anoadragon453, line 35: Looks like Jack and I never explained exactly *why* we formatted attachment's local filepaths in this manner (with the message ID and other bits of text). This docstring would be a handy place to do so.

```suggestion
    """Generate a unique filepath for a given attachment for the purposes of saving it to disk.
    
    Generates a filepath in the following format:
    
    <attachment_directory>/<Discord message ID>.<attachment ID>.<file extension>
    
    For example:
    
    /home/user/busty/625891304148303894.625891304081063986.mp3
    
    Filepaths for each song should be unique and the rest of the codebase assumes that they are.
    A Discord message can contain multiple attachments, including the attachment ID as well.
    
    Args:
    	message_id: The ID of the Discord message that included this attachment.
    	attachment: The attachment to build a filepath for.
    	
    Returns:
    	An absolute filepath to a file for this attachment in the configured attachment directory.
    """
```

Although now after having written this up, I'm wondering if we can just get away with using attachment IDs in order to get a unique filename. Ah well, an issue for another PR :)

src/discord_utils.py
```diff
@@ -31,6 +31,26 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def attachment_local_filepath(message_id: int, attachment: Attachment) -> str:
```
anoadragon453, line 34: nit: I would call this `build_filepath_for_attachment`. `attachment_local_filepath` sounds like we're extracting a filepath that already exists.

src/main.py
```diff
@@ -237,6 +241,51 @@ async def info(interaction: Interaction) -> None:
...
+    submit_message: Optional[str] = SlashOption(
+        required=False, description="The submission message text."
+    ),
+) -> None:
+    """Show a preview of a submission's "Now Playing" embed."""
+    attachment_filepath = discord_utils.attachment_local_filepath(
+        interaction.id, uploaded_file
+    )
+    random_emoji = random.choice(config.emoji_list)
+    await uploaded_file.save(fp=attachment_filepath)
```
anoadragon453, line 258: I would wait until after we've determined the file is valid media before saving it to disk.

anoadragon453: Also, could you leave a comment explaining why we need to save the file as well, if we're only going to throw it away later? (I know why, it's just not obvious at a glance :)

src/main.py
```diff
@@ -237,6 +241,51 @@ async def info(interaction: Interaction) -> None:
...
+
+    if cover_art is not None:
+        embed.set_image(url=f"attachment://{cover_art.filename}")
+        await interaction.response.send_message(
+            file=cover_art, embed=embed, ephemeral=True
+        )
+
+    else:
+        await interaction.response.send_message(embed=embed, ephemeral=True)
+    os.remove(attachment_filepath)
```
anoadragon453, line 286:
```suggestion
        await interaction.response.send_message(embed=embed, ephemeral=True)

    os.remove(attachment_filepath)
```

src/song_utils.py
```diff
@@ -8,12 +8,45 @@
...
+def embed_song(
+    message_content: str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User,
+    emoji: str,
+    jump_url: str,
+) -> Embed:
+
+    # Build and send "Now Playing" embed
```
anoadragon453, line 27: I think this comment might be outdated - `embed_song` doesn't send the embed, it only returns it.

src/config.py
```diff
@@ -27,6 +27,8 @@
 VOLUME_MULTIPLIER = 0.5
 # The maximum number of songs to download concurrently
 MAXIMUM_CONCURRENT_DOWNLOADS = 8
+# Holds a default URL value jump url for jump_url parameter in embed_song.
+PREVIEW_JUMP_URL = "https://youtu.be/J45GvH2_Ato"
```
anoadragon453, line 31: This URL never gets overridden, so I wouldn't call it "default".

```suggestion
# The URL that the '↲jump' link will lead to when using the /preview command.
PREVIEW_JUMP_URL = "https://youtu.be/J45GvH2_Ato"
```

anoadragon453 (changes requested):
This is looking great! A number of comments and suggestions below, but otherwise lgtm.

src/discord_utils.py
Cephian: Yeah just using attachment ID is probably a better solution. Or better yet, use <server_id>/<attachment_id>.<file_ext> to move us closer to server-independence. But I agree we'll do this in a separate PR

README.md
```diff
@@ -103,25 +104,26 @@ It should connect to Discord and display the currently logged-in application nam
...
+- An admin can run `/skip` at any time to skip the current song, or `/stop` to manually stop the show.
+- Once the last song has played, the bot will post a concluding message and leave the call.
 
 ### Command Reference
 
-1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `/bust`. 
+1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `/bust`.
 1. `/bust [<song #>]` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work. Specifying a song index will skip to that index before playing.
 1. `/info` - Show some statistics about currently listed songs.
+1. `/preview [<attachment>] [<submit_message>]` - Show a preview of a file's artist, title and artwork.
```
Cephian, line 120: There should probably be some indication that [<submit_message>] is optional like in the description for the /list command, though maybe we should eventually add a different notation for this

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def build_filepath_for_attachment(message_id: int, attachment: Attachment) -> str:
+    """
+    Generate a unique filepath for a given attachment for the purposes of saving it to disk
+
+    Generates a filepath in the following format:
```
Cephian, line 38: imo all these lines specifying the exact format are unnecessary to know how to use the function and just take up comment space, especially considering the code right below which actually does this is fairly readable. The comment should be for stuff that's not so understandable directly from code below.

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+    <attachment_directory>/<Discord message ID>.<attachment ID>.<file extension>
+
+    For example:
+    /home/user/busty/625891304148303894.625891304081063986.mp3
+
+    Filepaths for each song should be unique and the rest of the codebase assumes that they are.
```
Cephian, line 44: nitpick: filepaths for each _attachment_ should be unique

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+    A Discord message can contain multiple attachments, including the attachment ID as well.
```
Cephian, line 45: I'm not sure this sentence adds anything or even what the part after the comma is trying to say?

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def build_filepath_for_attachment(message_id: int, attachment: Attachment) -> str:
+    """
+    Generate a unique filepath for a given attachment for the purposes of saving it to disk
```
Cephian, line 36:
```suggestion
    Generate a unique filepath for a given attachment for the purposes of saving it to disk.
```
;)

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+    /home/user/busty/625891304148303894.625891304081063986.mp3
+
+    Filepaths for each song should be unique and the rest of the codebase assumes that they are.
+    A Discord message can contain multiple attachments, including the attachment ID as well.
+
+    Args:
+        message_id: The ID of the Discord message that included this attachment.
+        attachment: The attachment to build a filepath for.
+
+    Returns:
```
Cephian, line 51: maybe just personal preference but I feel like what it returns is implied by the description, docstrings are probably better short and sweet for smth like Busty

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+            os.path.splitext(attachment.filename)[1],
+        ),
+    )
+    return filepath
+
+
+def is_valid_media(attachment_content_type: Optional[str]) -> bool:
+    """Returns whether an attachment's content type is considered "media".
+
+    Valid media types are those that start with "audio" or "video".
```
Cephian, line 68: Again this is not hard to see from the code below, and maybe too specific to the function's defn in case we change it later

src/discord_utils.py
```diff
@@ -31,6 +31,54 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
+
+    Args:
+        attachment_content_type: A content type (aka MIME type).
+
+    Returns:
```
Cephian, line 73: again personal preference, but I think a separate returns section here just clutters the code with comments implied by other sections

src/song_utils.py
```diff
@@ -8,12 +8,45 @@
...
+def embed_song(
+    message_content: str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User,
+    emoji: str,
+    jump_url: str,
+) -> Embed:
+
+    # Build and return "Now Playing" embed
```
Cephian, line 27: Maybe make this into a docstring, the text is fine just surround with """

Cephian (changes requested):

README.md
anoadragon453: In the notation we currently have (and it standard across cli tools to the best of my knowledge) `[]` indicate an optional argument.

So in fact in this case the surrounding `[]`'s should be removed from `<attachment>`, since it is a required argument.

Cephian: In our current notation we seem to just use `[<arg>]` for all arguments, though I agree that the `[]` should be removed for non-optional ones. See the `\image url <url>` command etc where `url` is not optional. I also just noticed that the `\image view` and `\image clear` commands are listed as having arguments, which is just false lol

src/discord_utils.py
Cephian: This comment still applies lol

src/main.py
```diff
@@ -237,6 +241,54 @@ async def info(interaction: Interaction) -> None:
...
+    submit_message: Optional[str] = SlashOption(
+        required=False, description="The submission message text."
+    ),
+) -> None:
+    """Show a preview of a submission's "Now Playing" embed."""
+    user = interaction.user
+
+    if not discord_utils.is_valid_media(uploaded_file.content_type):
+        await interaction.response.send_message(
+            "You didn't send a valid media type. \nTry again.",
```
Cephian, line 258: Nitpick, but space and then newline is a little weird. I'd probably prefer "You didn't send a valid media type, try again." anyways.

src/main.py
```diff
@@ -237,6 +241,54 @@ async def info(interaction: Interaction) -> None:
...
+            ephemeral=True,
+        )
+        return
+
+    attachment_filepath = discord_utils.build_filepath_for_attachment(
+        interaction.id, uploaded_file
+    )
+    random_emoji = random.choice(config.emoji_list)
+    await uploaded_file.save(fp=attachment_filepath)
```
Cephian, line 267:
```suggestion
    await uploaded_file.save(fp=attachment_filepath)
    random_emoji = random.choice(config.emoji_list)
```
nitpick, but i'd swap these lines since random_emoji is related to the line after this block, and the uploaded file save relates to the lines above this block.

Cephian (changes requested):

src/discord_utils.py
Cephian: I see it doesn't necessarily show the previous comment in the timeline so I'll repaste it

> imo all these lines specifying the exact format are unnecessary to know how to use the function and just take up comment space, especially considering the code right below which actually does this is fairly readable. The comment should be for stuff that's not so understandable directly from code below.

README.md
```diff
@@ -103,25 +104,26 @@ It should connect to Discord and display the currently logged-in application nam
...
+1. `/list [<channel>]` - Download and list all media sent in the current text channel. Specifying a channel will cause songs to be pulled from that channel instead. This must be run before `/bust`.
+1. `/bust [<song_num>]` - Join the vc/stage that the user who ran this command is currently in, and plays the tracks in the channel in order. The user must be in a vc or stage for this to work. Specifying a song index will skip to that index before playing.
 1. `/info` - Show some statistics about currently listed songs.
-1. `/image upload [<attachment>]` - Upload an image to be used in the Google Form generated when running `/list`.
-1. `/image url [<url>]` - Queue an image to be used in the Google Form generated when running `/list`.
-1. `/image view [<url>]` - View currently loaded image to be used in the Google Form generated when running `/list`.
-1. `/image clear [<url>]` - Clear currently loaded image to be used in the Google Form generated when running `/list`.
-1. `/skip` - Skips the current track :scream: 
-1. `/stop` - Stop busting early :scream: :scream: :scream: 
+1. `/preview <attachment> [<submit_message>]` - Show a preview of a file's artist, title and artwork.
```
Cephian, line 120:
```suggestion
1. `/preview <attachment> [<submit_message>]` - Show a preview of the attachment's artist, title and artwork.
```

Cephian (changes requested):
My only comment is just a wording nitpick, other than that this is an approval from me. Not marking this review as "Approve" only bc I want to wait for @anoadragon453's approval as well

Cephian (commented):
LGTM, will wait to see what @anoadragon453 says

src/discord_utils.py
```diff
@@ -31,6 +31,27 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def build_filepath_for_attachment(message_id: int, attachment: Attachment) -> str:
+    """Generate a unique filepath for a given attachment for the purposes of saving it to disk."""
```
anoadragon453, line 35: Did you mean to only include the first line of the descriptions in https://github.com/anoadragon453/busty/pull/167#discussion_r1103884933 and https://github.com/anoadragon453/busty/pull/167#discussion_r1103887258?

src/main.py
```diff
@@ -237,6 +241,56 @@ async def info(interaction: Interaction) -> None:
     await bc.send_stats(interaction)
 
 
+# Preview command
+@client.slash_command(dm_permission=False)
+async def preview(
+    interaction: Interaction,
+    uploaded_file: Attachment = SlashOption(description="The file to submit."),
```
anoadragon453, line 248:
```suggestion
    uploaded_file: Attachment = SlashOption(description="The song to submit."),
```

Just so folks don't think cover art needs to be put here, or something.

src/main.py
```diff
@@ -237,6 +241,56 @@ async def info(interaction: Interaction) -> None:
...
+    submit_message: Optional[str] = SlashOption(
+        required=False, description="The submission message text."
+    ),
+) -> None:
+    """Show a preview of a submission's "Now Playing" embed."""
+    await interaction.response.defer(ephemeral=True)
+    user = interaction.user
```
anoadragon453, line 255: I'm not sure if this is really that helpful, especially when `user` is only used once below.

src/song_utils.py
```diff
@@ -8,12 +8,45 @@
...
 
+def embed_song(
+    message_content: str,
+    attachment_filepath: str,
+    attachment: Attachment,
+    user: User,
+    emoji: str,
+    jump_url: str,
+) -> Embed:
+    """Build and return "Now Playing" embed"""
```
anoadragon453, line 26:
```suggestion
    """Build and return a "Now Playing" embed"""
```

anoadragon453 (changes requested):
Only a few notes, otherwise LGTM!

src/discord_utils.py
jack-cogan: Yeah, I agreed with Jack on specifying the exact format is unnecessary and just takes up space.
https://github.com/anoadragon453/busty/pull/167#discussion_r1106192333

jcogan97: Yeah, I agreed with Jack on specifying the exact format of this function was unnecessary and just takes up space, so I've left it to describe what the function does when given an attachment and message id.
https://github.com/anoadragon453/busty/pull/167#discussion_r1106192333

src/main.py
Cephian: Have you tested this since adding the defer to the beginning? It may have timed out if the attachment save took longer than 3 seconds without a defer.

It doesn't make sense looking at this that it's necessary. In general it's a good idea to find the reason for unexpected behavior like this instead of doing something less optional that seems to work

Cephian (approved):

src/discord_utils.py
anoadragon453: Hum, I still think useful information from those comments is being left out, but I agree that at least the descriptions of the arguments and return types are sometimes redundant. In that case, we can leave `build_filepath_for_attachment` alone.

For `build_filepath_for_attachment`, it's useful for callers to know that the returned filepath will be absolute and located in the attachment dir, so we should at least add that bit?

```suggestion
    """Generate a unique, absolute filepath for a given attachment located in the configured attachment directory."""
```

As for the description of the makeup of the filepath, we should still describe *why* we chose that as the format. Doing so helped us understand that our logic was in fact flawed, and we didn't need to include the Discord message ID in order to get a unique filepath.

That all may be better left to a comment in the method though (i.e. below), as we know calling functions in our codebase won't parse this filepath. If this were a public function in a library, I'd be more inclined to include the below in the docstring, for the caller may end up needing to know the structure of the returned path (in order to extract part of it).

```py
	# Generate and return a filepath in the following format:
    #     <attachment_directory>/<Discord message ID>.<attachment ID>.<file extension>
    # For example:
    #     /home/user/busty/625891304148303894.625891304081063986.mp3
    # 
    # The Discord Message ID is included as well, as individual messages can have multiple attachments.
    filepath = path.join(
        config.attachment_directory_filepath,
        "{}.{}{}".format(
            message_id,
            attachment.id,
            os.path.splitext(attachment.filename)[1],
        ),
    )
    return filepath
```

In general I tend to over-document rather than under-document. It can become especially useful when your codebase becomes larger and older, and you're left having to crawl through PRs or message people in order to understand why certain choices were made and whether changing code will break something.

src/discord_utils.py
```diff
@@ -31,6 +31,35 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
...
 
+def build_filepath_for_attachment(message_id: int, attachment: Attachment) -> str:
+    """Generate a unique, absolute filepath for a given attachment located in the configured attachment directory."""
+
+    # Generate and return a filepath in the following format:
+    #     <attachment_directory>/<Discord message ID>.<attachment ID>.<file extension>
+    # For example:
+    #     /home/user/busty/625891304148303894.625891304081063986.mp3
+
+    # The Discord Message ID is included as well, as individual messages can have multiple attachments.\
```
anoadragon453, line 42:
```suggestion
    # The Discord Message ID is included as well, as individual messages can have multiple attachments.
```

anoadragon453 (approved):

PR #168, opened 2023-02-08
Return Unix Timestamp inside /info embed object

### LINKED ISSUE #151: Add 'Ending time of bust' to /info command
Once #150 merges, we should add a field which shows up in the /info command during a bust, which sends a Discord timestamp to the ending time of the bust.

Thanks to @anoadragon453 for the idea.

jcogan97:
Closes #151.

`/info` will now return a timestamp recording the time in which the current bust ends.

Returns:
CleanShot 2023-02-08 at 23 51 14@2x

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -1,5 +1,7 @@
 import asyncio
+import datetime
 import random
+import time
 from collections import defaultdict
 from typing import List, Optional, Tuple, Union
 
@@ -293,10 +295,7 @@ class BustController:
             for submit_message, attachment, local_filepath in self.bust_content
         ]
 
-        # Extract bust number from channel name
-        bust_number = "".join([c for c in self.message_channel.name if c.isdigit()])
-        if bust_number:
-            bust_number = bust_number + " "
+        bust_number = discord_utils.extract_bust_number(self.message_channel)
 
         form_url = forms.create_remote_form(
             f"Busty's {bust_number}Voting",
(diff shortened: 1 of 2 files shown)


anoadragon453:
@jcogan97 Is this ready for review? If not, could you mark it as a draft again?

Also heads up that you've got a conflicting file.

src/bust.py
```diff
@@ -314,7 +316,18 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
     async def send_stats(self, interaction: Interaction) -> None:
         songs_len = int(self.total_song_len)
         num_songs = len(self.bust_content)
-        bust_len = songs_len + config.seconds_between_songs * num_songs
+        bust_len_in_seconds = songs_len + config.seconds_between_songs * num_songs
```
Cephian, line 319: instead of calculating both `bust_len_in_seconds` and `remaining_bust_len_in_seconds` and just showing one based on the if statement, you should probably just have one variable which branches based on whether we're busting or not

src/discord_utils.py
```diff
@@ -31,6 +31,15 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
+def extract_bust_number(message_channel: TextChannel) -> str:
```
Cephian, line 34: Is this refactor related to anything? It seems like this is only called once

src/bust.py
```diff
@@ -314,7 +316,18 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
...
+
+        remaining_bust_len = sum(
```
Cephian, line 321: This fails to take into account how far we are in the currently playing song

Cephian (changes requested):
As a side note, I'll also mention that we don't really know how much things like API latency and Python execution time play into this, so it will definitely need to be tested with wall-clock-time to see how accurate bust length estimates are.

PR #170, opened 2023-02-11
Remove in-text emoji, use discord emoji format

Cephian:
Just some nice cleanup, we use discord emoji format in emoji_list.py and it was a bit ugly having in-text emoji mixed into an otherwise ascii codebase

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -165,7 +165,8 @@ class BustController:
             await bot_member.edit(nick=self.original_bot_nickname)
 
         if say_goodbye:
-            embed_title = "❤️‍🔥 That's it everyone ❤️‍🔥"
+            goodbye_emoji = ":heart_on_fire:"
+            embed_title = f"{goodbye_emoji} That's it everyone {goodbye_emoji}"
             embed_content = "Hope ya had a good **BUST!**"
             embed_content += "\n*Total length of all submissions: {}*".format(
                 song_utils.format_time(int(self.total_song_len))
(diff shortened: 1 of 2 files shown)


src/bust.py
```diff
@@ -165,7 +165,8 @@ async def finish(self, say_goodbye: bool = True) -> None:
             await bot_member.edit(nick=self.original_bot_nickname)
 
         if say_goodbye:
-            embed_title = "❤️‍🔥 That's it everyone ❤️‍🔥"
+            goodbye_emoji = ":heart_on_fire:"
+            embed_title = f"{goodbye_emoji} That's it everyone {goodbye_emoji}"
```
anoadragon453, line 169: I'm not sure if separating `:heart_on_fire:` into a separate variable buys us much here?

src/bust.py
```diff
@@ -364,14 +365,15 @@ async def create_controller(
...
-            content="\N{WARNING SIGN} No valid media files found."
+            content=":warning: No valid media files found."
         )
         return None
 
     bc = BustController(client, channel_media_attachments, interaction.channel)
 
     # Title of /list embed
-    embed_title = "❤️‍🔥 AIGHT. IT'S BUSTY TIME ❤️‍🔥"
+    bust_emoji = ":heart_on_fire:"
```
anoadragon453, line 375: Likewise not sure if this is needed.

anoadragon453 (approved):
Nice change! I didn't realise we had so much inconsistency in the way we name emoji.

src/bust.py
Cephian: It doesn't save us any characters, my reasoning was that it visually separates it from the text in syntax highlighting, and that if we ever theoretically wanted to change it it would be in one location. It's probably pretty arbitrary which is better though

PR #172, opened 2023-02-15
Defer potentially long commands

### LINKED ISSUE #171: The bot can fail to play songs when Discord's API is slow
We were consistently seeing application response errors when calling `/bust` on the Valentine's Day bust:

image

We figured this must be related to Discord's unfortunately timed jump in API response times:

image

We were also seeing the following in busty's logs:

```
Command raised an exception: NotFound: 404 Not Found (error code: 10062): Unknown interaction
```

It appears that Discord has a 3s time limit for bots to respond to interactions: https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-modal. If we don't respond in time, the interaction will no longer be recognised by Discord, and responding to the interaction will raise a `NotFound` exception on the client. That would be raised here:

https://github.com/anoadragon453/busty/blob/426ee5ff8df58a99bb9fb6e51b467de6e1d6e2ac/src/bust.py#L132

Which would exit the `play` method, preventing busty from actually playing any songs. There are some API calls above which involve getting voice/stage channels and member information for the guild, as well as joining the voice call. All of that involves multiple API calls and with each taking ~100ms+ that may have added up to >3s.

To get around this, we can initially acknowledge the response (an [interaction callback type](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-interaction-callback-type) of PONG), and then send [followup messages](https://discord.com/developers/docs/interactions/receiving-and-responding#followup-messages), which would be our actual messages. That would extend our time limit from 3s to 15m.

Cephian:
Closes #171.  Any commands which could potentially take a long time* are now deferred. Also, I've refactored so that when an interaction is passed to a bust controller function it can be assumed to be handled "within" and that bust controller function never returns with an non-replied-to interaction. Defer, error-handling, and "OK" interaction replies are now in the bust controller.

* Technically when saving persistent states file write could take a very long time, but I didn't defer those. Also I guess your CPU clock could be reeeeeally slow....

When using `interaction.response.defer()` your one initial interaction allowance is spent, so you have to use `interaction.followup.send` instead of `interaction.response.send()` afterwards.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -85,6 +85,7 @@ class BustController:
             self.play_song_task.cancel()
 
     async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
+        await interaction.response.defer(ephemeral=True)
         # Join active voice call
         voice_channels: List[Union[VoiceChannel, StageChannel]] = list(
             interaction.guild.voice_channels
@@ -131,6 +132,7 @@ class BustController:
         self.original_bot_nickname = bot_member.display_name
 
         await interaction.channel.send("Let's get **BUSTY**.")
+        interaction.delete_original_message()
 
         # Play songs
         for index in range(skip_count, len(self.bust_content)):
@@ -311,6 +313,7 @@ class BustController:
         return form_url
 
     async def send_stats(self, interaction: Interaction) -> None:
+        interaction.defer(ephemeral=True)
         songs_len = int(self.total_song_len)
         num_songs = len(self.bust_content)
         bust_len = songs_len + config.seconds_between_songs * num_songs
@@ -360,6 +363,7 @@ async def create_controller(
     list_channel: TextChannel,
 ) -> Optional[BustController]:
     """Attempt to create a BustController given a channel list command"""
+    await interaction.response.defer(ephemeral=True)
     # Scrape all tracks in the target channel and list them
     channel_media_attachments = await discord_utils.scrape_channel_media(list_channel)
     if not channel_media_attachments:
@@ -453,5 +457,6 @@ async def create_controller(
             )
             await discord_utils.try_set_pin(form_message, True)
 
+    await interaction.delete_original_message()
     # Return controller
     return bc
--- a/src/bust.py
+++ b/src/bust.py
@@ -93,7 +93,7 @@ class BustController:
         voice_channels.extend(interaction.guild.stage_channels)
 
         if not voice_channels:
-            await interaction.response.send_message(
+            await interaction.followup.send(
                 "You need to be in an active voice or stage channel.", ephemeral=True
             )
             return
@@ -122,7 +122,7 @@ class BustController:
                 return
         else:
             # No voice channel was found
-            await interaction.response.send_message(
+            await interaction.followup.send(
                 "You need to be in an active voice channel.", ephemeral=True
             )
             return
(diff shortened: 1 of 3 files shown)


src/bust.py
```diff
@@ -311,6 +313,7 @@ def get_google_form_url(self, image_url: Optional[str] = None) -> Optional[str]:
         return form_url
 
     async def send_stats(self, interaction: Interaction) -> None:
+        interaction.defer(ephemeral=True)
```
jack-cogan, line 316: This is just for my own understanding, but why are we not awaiting this defer like for `play` and `create_controller`?

src/bust.py
```diff
@@ -131,6 +132,7 @@ async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
         self.original_bot_nickname = bot_member.display_name
 
         await interaction.channel.send("Let's get **BUSTY**.")
+        interaction.delete_original_message()
```
jcogan97, line 135: This is just for my own understanding, but why are we not awaiting this defer like for `play` and `create_controller`?

src/bust.py
jcogan97, line 316: This is just for my own understanding, but why are we not awaiting this defer like for `play` and `create_controller`?

Cephian: bc I forgot to write await lol, will fix later today

anoadragon453 (changes requested):

src/bust.py
```diff
@@ -455,5 +459,6 @@ async def create_controller(
             )
             await discord_utils.try_set_pin(form_message, True)
 
+    await interaction.delete_original_message()
```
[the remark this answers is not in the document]
Cephian: A part of why I felt it was better to put them inside the BustController is that error messages are send in the bust controller, and it seemed cleaner for a function to return a bust controller where the interaction is definitely handled internally (either by error, by deleting command, by responding ok, etc), instead of one where whether or not it's been handled relies on whether or not the function succeeds.

The most proper way to really decouple all this IMO would be to return several different error codes and handle all those messages outside of the BustController, but that might be overengineering for our use case

src/bust.py
```diff
@@ -85,14 +85,15 @@ def skip_song(self) -> None:
             self.play_song_task.cancel()
 
     async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
+        await interaction.response.defer(ephemeral=True)
```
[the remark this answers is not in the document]
Cephian: I do agree that conceptually direct slash command handling is outside the bust controller, as mentioned in my other comment I think it's sort of a side effect of us handling errors in the BustController directly at the interaction level that it breaks some encapsulation by deferring outside, responding to errors inside, then having to check for errors after it returns etc.

My intention was to make the bustcontroller functions guarantee that they 1) take an interaction not responded to in any way 2) upon return, guarantee that interaction has been "dealt with".

As mentioned in my other comment, I think the "correct" abstraction is to allow returning some error codes/throwing some exceptions and handle all messages upon return outside the controller. But this might be just complicating things for no reason in our case. lmk what you think

anoadragon453 (approved):
After chatting internally over the model of bustcontroller methods expecting non-replied to Interaction objects, this looks good to me.

My one wish would be to add a line to each relevant BustController method's docstring stating that the function assumes that the passed `Interaction` object has not yet been responded to.

PR #175, opened 2023-03-05
Add multiserver support

### LINKED ISSUE #148: Allow usage on multiple servers
While it's admittedly unlikely we'll ever have a single busty instance being used by two servers at a time, I think this a worthwhile upgrade mainly so that we can actually test production busty after updating the server. 

I believe the main two things that need to be done for this are
1) Turn global saved image into map guild -> saved image
2) Turn global attachments directory into map guild -> attachments dir

Cephian:
Closes #148.

The main changes are
1) Turn attachments directory format in `build_filepath_for_attachment()` from `<attachment_dir>/<message_id>.<attachment_id>.<extension>` to `<attachment_dir>/<guild_id>/<attachment_id>.<extension>`.
2) Move clearing useless cached files above the code for downloading new files, to save on disk space
3) In `main.py`, remove the `get_controller()` method, which was a hack since we could not previously have a BustController remove itself from the guild_id --> BustController map after busting. Now that BustController.play() returns only after the bust is complete we can do it ourselves.

Thankfully the persistent_state module was written with multiserver in mind, so no changes needed to be made there. This has all been tested to the best of my abilities.

### DIFF
--- a/src/discord_utils.py
+++ b/src/discord_utils.py
@@ -31,20 +31,18 @@ async def try_set_pin(message: Message, pin_state: bool) -> None:
         print("Altering message pin state failed:", e)
 
 
-def build_filepath_for_attachment(message_id: int, attachment: Attachment) -> str:
+def build_filepath_for_attachment(guild_id: int, attachment: Attachment) -> str:
     """Generate a unique, absolute filepath for a given attachment located in the configured attachment directory."""
 
     # Generate and return a filepath in the following format:
-    #     <attachment_directory>/<Discord message ID>.<attachment ID>.<file extension>
+    #     <attachment_directory>/<Discord guild ID>/<attachment ID>.<file extension>
     # For example:
-    #     /home/user/busty/625891304148303894.625891304081063986.mp3
-
-    # The Discord Message ID is included as well, as individual messages can have multiple attachments.
+    #     /home/user/busty/attachments/922994022916698154/625891304081063986.mp3
 
     filepath = path.join(
         config.attachment_directory_filepath,
-        "{}.{}{}".format(
-            message_id,
+        str(guild_id),
+        "{}{}".format(
             attachment.id,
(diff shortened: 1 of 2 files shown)


src/discord_utils.py
```diff
@@ -66,9 +64,13 @@ async def scrape_channel_media(
...
 
+    attachment_dir = path.join(
+        config.attachment_directory_filepath, str(channel.guild.id)
+    )
+
     # Ensure attachment directory exists
-    if not os.path.exists(config.attachment_directory_filepath):
-        os.mkdir(config.attachment_directory_filepath)
+    if not os.path.exists(attachment_dir):
+        os.makedirs(attachment_dir)
```
anoadragon453, line 73: Ah, arguably we should have made this `os.makedirs` to begin with!

src/discord_utils.py
```diff
@@ -93,6 +97,15 @@ async def scrape_channel_media(
                 )
             )
 
+    # Clear unused files in attachment directory
```
anoadragon453, line 100:
```suggestion
    # Clear unused files in this guild's attachment directory
```

Just so it doesn't sound like we're deleting any attachments related to other guilds.

src/main.py
```diff
@@ -70,7 +58,8 @@ async def on_list(
     ),
 ) -> None:
     """Download and list all media sent in a chosen text channel."""
-    bc = get_controller(interaction.guild_id)
+    global controllers
```
anoadragon453, line 61: One only needs to do `global X` if they're changing the object that `X` points to. In this function we're only modifying what `controllers` contains, not actually re-assigning it (`controllers =`).

So the existing `global controller` and this one should be unnecessary.

Likewise for the other functions.

anoadragon453 (approved):
A couple changes, otherwise LGTM!

src/main.py
Cephian: TIL

PR #176, opened 2023-03-10
Fix issues with Mutagen file detection

### LINKED ISSUE #162: Correctly detect the length of songs with incorrect file extensions
Due to some interesting filetype detection algorithms in mutagen, Busty does not always correctly read tags from files with incorrect file extensions, even though she will still play them fine with ffmpeg.

One example is .aac/.m4a files given the .wav extension, mutagen will detect them as wav. and fail to parse file length.

Right now Busty is running on the branch `ugly_info_fix` which retries failed parsing of song lengths by symlinking the same file under lots of different extensions and hoping one of them works. This is probably not the best solution. Paths forward I can see are

1) Fix in upstream mutagen.for the m4a/wav example this might be an easy fix, though in general their file detection is somewhat questionable for this mismatched codec/extension case.
2) Use an external magic byte/file detection library and more intelligently symlink and decode
3) Merge the ugly info fix, either permanently or until we have something better

Cephian:
Closes #162, where Mutagen overweights file extension in guessing filetypes. With this we can finally retire the `ugly_info_fix` branch.

There are basically three solutions to this problem that I can think of.

1. Solve the problem upstream in Mutagen. The most morally righteous, but would may require navigating some pushback and take time to get merged and released
2. Detect better file extension (with magic/guessing/using default ext with no hints), symlink using that extension, and pass the symlink to mutagen. Variants for doing this every time, only if reading fails, etc etc. All of these are a little ugly and hacky in my opinion.
3. ⭐ What this PR does, remove file extensions when we save attachments in the first place. FFMPEG should have no issue determining file formats without an extension and Mutagen seems not to either in my testing, though if this is merged it would be good to test all past busts to be sure.

IMO 3 is the best solution because it seems like something we could have reasonably done intentionally as a design decision in the first place instead of to circumvent a bug. Let me know your thoughts.

This PR also contains very minor formatting changes in `song_utils.py` that don't change any logic, just 'cause.

### DIFF
--- a/src/discord_utils.py
+++ b/src/discord_utils.py
@@ -35,17 +35,12 @@ def build_filepath_for_attachment(guild_id: int, attachment: Attachment) -> str:
     """Generate a unique, absolute filepath for a given attachment located in the configured attachment directory."""
 
     # Generate and return a filepath in the following format:
-    #     <attachment_directory>/<Discord guild ID>/<attachment ID>.<file extension>
+    #     <attachment_directory>/<Discord guild ID>/<attachment ID>
     # For example:
-    #     /home/user/busty/attachments/922994022916698154/625891304081063986.mp3
+    #     /home/user/busty/attachments/922994022916698154/625891304081063986
 
     filepath = path.join(
-        config.attachment_directory_filepath,
-        str(guild_id),
-        "{}{}".format(
-            attachment.id,
-            os.path.splitext(attachment.filename)[1],
-        ),
+        config.attachment_directory_filepath, str(guild_id), f"{attachment.id}"
     )
     return filepath
 
--- a/src/discord_utils.py
+++ b/src/discord_utils.py
@@ -40,7 +40,7 @@ def build_filepath_for_attachment(guild_id: int, attachment: Attachment) -> str:
     #     /home/user/busty/attachments/922994022916698154/625891304081063986
 
     return path.join(
-        config.attachment_directory_filepath, str(guild_id), f"{attachment.id}"
+        config.attachment_directory_filepath, str(guild_id), str(attachment.id)
     )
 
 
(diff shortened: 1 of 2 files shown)


src/discord_utils.py
```diff
@@ -35,17 +35,12 @@ def build_filepath_for_attachment(guild_id: int, attachment: Attachment) -> str:
...
+    #     /home/user/busty/attachments/922994022916698154/625891304081063986
 
     filepath = path.join(
-        config.attachment_directory_filepath,
-        str(guild_id),
-        "{}{}".format(
-            attachment.id,
-            os.path.splitext(attachment.filename)[1],
-        ),
+        config.attachment_directory_filepath, str(guild_id), f"{attachment.id}"
```
anoadragon453, line 43: nit: It's a bit odd that we're using both `str()` and `f"{}"` as int -> str conversion methods here. Let's just pick one?

anoadragon453 (approved):
Yes, I don't see much harm in removing the file extension from the attachments in the attachment directory. We may just want to look over the filetypes that mutagen accepts and see if this will affect their score values negatively in any way.

lgtm otherwise, and one nit below.

Cephian:
> We may just want to look over the filetypes that mutagen accepts and see if this will affect their score values negatively in any way.

This [looks true](https://github.com/search?q=repo%3Aquodlibet%2Fmutagen%20score\(&type=code) at least for the main formats MP3, OGG, FLAC, WAV. It also seems true for most others. The only weird case I see skimming is that it seems [AAC is based only on filename cues](https://github.com/quodlibet/mutagen/blob/83e1169d52b7f48bf0cc9eec3bc8ee736e96d9ee/mutagen/aac.py#L412). This seems fine to me, but lmk how you feel about it @anoadragon453.

PR #177, opened 2023-03-13
Simplify interaction response

Cephian:
It turns out that instead of using `interaction.response.send_message()` or `interaction.followup.send()` based off of whether we have deferred or not, we can just use `interaction.send()` and nextcord calls the correct one for us. Not sure how we missed this one lol.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -98,7 +98,7 @@ class BustController:
         voice_channels.extend(interaction.guild.stage_channels)
 
         if not voice_channels:
-            await interaction.followup.send(
+            await interaction.send(
                 "You need to be in an active voice or stage channel.", ephemeral=True
             )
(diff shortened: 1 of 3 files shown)


jcogan97:
Do we still need to defer to acknowledge an interaction, if `interaction.send()` handles the followup?

Other than that, LGTM

Cephian:
> Do we still need to defer to acknowledge an interaction, if `interaction.send()` handles the followup?
> 
> Other than that, LGTM

Yes, the point of deferral is to 1) show the user Busty is thinking before long operations, and 2) ack the interaction so it doesn't timeout in 3 seconds. This code is functionally equivalent to what existed before, it just chooses response vs followup based on whether the interaction has been responded to already.

Issue #179, opened 2023-03-14
Google Forms cannot be viewed until an authorised google user attempts to edit it


anoadragon453:
A strange behaviour has cropped up with our Google Forms. After the form is generated, a "responder link" is sent in an embed to the channel for people to click and vote on songs with.

Two things happened starting this week that interrupted this process:

- Using @Cephian's google service account credentials, the form created by the bot could not be viewed by everyone. You would receive a Permission denied error related to organisation permissions ([example form link](https://docs.google.com/forms/d/e/1FAIpQLSccLWVdPFN2UZ82vWH2GzdO5nb-JCKw4Nbxma_NwMAM_zpERQ/formrestricted)):

![image](https://user-images.githubusercontent.com/1342360/225143876-aa0d5246-9b0d-4e1e-a4d7-265c7d636133.png)

If you try to access the form as an anonymous user, you get the sign-in page. If you try to access it as a registered user (even as our accounts) you get that permissions error.

To work around this for Busty's 30, we switched to my google services credentials.

- The new problem was that the form was created, but inaccessible to anyone *until* I (and possibly anyone else on the Busty's team with google auth) try to edit the form. After you open the form in google drive, anyone can access the form. Before you do that though, non-logged in users will again receive a sign in page ([example form](https://docs.google.com/forms/d/e/1FAIpQLSdOCeA-5LS6rCo23WVsZOb0W9GT9_N0jjWS9_q4GEcESSwgGA/viewform)).

We mused about having an admin quickly copy the form URL, edit it to replace `/viewform` with `/edit`, and possibly hiding the channel while that's done. Unfortunately, the `/edit` version of the URL uses a completely different ID to the "responder url" (`/edit` links use the form ID, responder urls use a separate identifier in the URL).

Instead, to work around this (temporary?) problem, we can just give everyone the `/edit` URL. Regular users will be directed to a sign-in page initially until an admin clicks that URL. Once they do, regular users will get the form as before (with a request edit access button in the corner, but that's fine).

Cephian:
Did we ever find out what caused this @anoadragon453? Can we close it?

anoadragon453:
@Cephian we did not; still assuming it was a change on Google's side that eventually got reverted?

I don't think this issue is helpful anymore as we're no longer experiencing it, so happy to close.

PR #180, opened 2023-03-14
Return the "edit" URL for google forms to work around #179

anoadragon453:
A temporary patch to work around https://github.com/anoadragon453/busty/issues/179. This implements the solution posited at the end of the issue:

> Instead, to work around this (temporary?) problem, we can just give everyone the /edit URL. Regular users will be directed to a sign-in page initially until an admin clicks that URL. Once they do, regular users will get the form as before (with a request edit access button in the corner, but that's fine).

**This PR should not be merged as-is.**

### DIFF
--- a/src/forms.py
+++ b/src/forms.py
@@ -141,4 +141,6 @@ def create_remote_form(
         except Exception as e:
             print("Error moving form:", e)
 
-    return form_url
+    #return form_url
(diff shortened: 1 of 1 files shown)


anoadragon453:
This isn't in use in production and we seem to be operating fine without it. It appears that it was just a workaround for a temporary change on Google's side.

Thus, there's no need to keep this PR open any longer.

PR #186, opened 2023-04-06
Make LLM bust-aware

Cephian:
Busty will now know when the bust is currently happening and the song name.

image

I moved the guild.id --> BustController map into `bust.py` so it can be accessed by anything that imports it, including `llm.py`. As a result I also had to rename the `bust()` function in `main.py` so the imported `bust` module could be accessed.

We could turn off her random responses during the bust message spam but I think it might be funnier if we test not doing that for this time

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -256,16 +256,14 @@ class BustController:
             after=ffmpeg_post_hook,
         )
 
-        now_playing_str = song_utils.song_format(
+        # Set now playing title
+        self.now_playing_str = song_utils.song_format(
             local_filepath, attachment.filename, submit_message.author.display_name
         )
 
-        # Set now playing title
-        self.now_playing_str = now_playing_str
-
         # Change the name of the bot to that of the currently playing song.
         # This allows people to quickly see which song is currently playing.
-        new_nick = random_emoji + now_playing_str
+        new_nick = random_emoji + self.now_playing_str
 
         # If necessary, truncate name to max length allowed by Discord,
         # appending an ellipsis on the end.
--- a/src/bust.py
+++ b/src/bust.py
@@ -1,7 +1,7 @@
 import asyncio
 import random
 from collections import defaultdict
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union, Dict
 
 from nextcord import (
     Attachment,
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -103,14 +101,14 @@ async def on_list(
...
+        bc = await bust.create_controller(client, interaction, list_channel)
+        bust.controllers[interaction.guild_id] = bc
 
 
 # Bust command
-@client.slash_command(dm_permission=False)
+@client.slash_command(name="bust", dm_permission=False)
 @application_checks.has_role(config.dj_role_name)
-async def bust(
+async def do_bust(
```
jcogan97, line 111: seems a little odd calling it `do_bust`, maybe call it something like `bust_action` or `launch_bust`

Cephian: probably on_bust since we have on_list, will change later

src/bust.py
```diff
@@ -250,12 +256,17 @@ async def ffmpeg_post_hook(e: Exception = None):
...
+        now_playing_str = song_utils.song_format(
             local_filepath, attachment.filename, submit_message.author.display_name
         )
 
+        # Set now playing title
+        self.now_playing_str = now_playing_str
+
+        # Change the name of the bot to that of the currently playing song.
+        # This allows people to quickly see which song is currently playing.
+        new_nick = random_emoji + now_playing_str
```
anoadragon453, line 268: You can combine these.

```suggestion
        # Set now playing title
        self.now_playing_str = song_utils.song_format(
            local_filepath, attachment.filename, submit_message.author.display_name
        )

        # Change the name of the bot to that of the currently playing song.
        # This allows people to quickly see which song is currently playing.
        new_nick = random_emoji + self.now_playing_str
```

src/llm.py
```diff
@@ -101,6 +102,15 @@ def get_random_context() -> List[str]:
...
 async def get_server_context(message: Message) -> List[str]:
     result = []
+
+    # Detect if song is currently playing
+    if hasattr(message, "guild"):
+        bc = bust.controllers.get(message.guild.id)
+        if bc and bc.is_active():
+            result.append("The bust is going on right now!")
+            result.append(f"Now playing: {bc.current_song()}")
+            result.append("Tell everyone you can't respond since you're busy busting.")
```
anoadragon453, line 112: I'm surprised this doesn't cause her automatic responses to be like "I'm busy" all the time.

anoadragon453 (approved):
lgtm, besides a couple minor points.

src/llm.py
Cephian: Yeah that's kind of what I was going for but I think what happened instead was way better

PR #190, opened 2023-06-15
Switch message channel to where bust command is issued from

Cephian:
We recently had an issue where Busty crashed mid-bust. This doesn't fix that, but it does fix another issue we found afterwards where if we listed from another channel (bod dev channel), it would continue to send messages from the channel we listed from, even if /bust was issued elsewhere (the mailbox).

This PR makes it so that the message channel (where messages are sent) gets updated whenever /list OR /bust is run.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -97,6 +97,7 @@ class BustController:
         """
 
         await interaction.response.defer(ephemeral=True)
+
         # Update message channel to where command was issued from
         self.message_channel = interaction.channel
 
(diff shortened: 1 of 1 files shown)


src/bust.py
```diff
@@ -97,6 +97,10 @@ async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
         """
 
         await interaction.response.defer(ephemeral=True)
+
+        # Update message channel to where command was issued from
```
anoadragon453, line 101:
```suggestion
        # Update message channel to where command was issued from
        # (in case `list` was called from a separate/private channel).
```

Just to say why we're doing this.

anoadragon453 (approved):
Apologies for the glacial review pace, I've been slackin' on busty.

LGTM, thank you!

PR #191, opened 2023-06-27
Make busty retry on "busy rn"

Cephian:
Make busty try one more time when she gets a timeout from openai before resorting to "busy rn"

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -31,6 +31,8 @@ MAXIMUM_CONCURRENT_DOWNLOADS = 8
 PREVIEW_JUMP_URL = "https://youtu.be/J45GvH2_Ato"
 # The OpenAI model to use for GPT abilities
 OPENAI_MODEL = "gpt-3.5-turbo"
+# Number of times to retry if API does not respond in time
+LLM_RETRY = 2
 
 # SETTINGS
 # How many seconds to wait in-between songs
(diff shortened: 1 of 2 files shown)


src/llm.py
```diff
@@ -291,15 +292,15 @@ async def get_response_text(message: Message) -> Optional[str]:
...
 
-    response = await query_api(data)
-    if response:
-        # Remove "Busty: " prefix
-        prefix = f"{get_name(self_user)}: "
-        if response.startswith(prefix):
-            response = response[len(prefix) :]
-
-        return response
+    for x in range(config.LLM_RETRY):
```
anoadragon453, line 295:
```suggestion
    for _ in range(config.LLM_RETRY):
```

If we're not using the variable, it's more common to use `_`.

src/config.py
```diff
@@ -31,6 +31,8 @@
 PREVIEW_JUMP_URL = "https://youtu.be/J45GvH2_Ato"
 # The OpenAI model to use for GPT abilities
 OPENAI_MODEL = "gpt-3.5-turbo"
+# Number of times to retry if API does not respond in time
+LLM_RETRY = 2
```
anoadragon453, line 35:
```suggestion
LLM_RETRY_COUNT = 2
```

may be slightly clearer.

anoadragon453 (approved):
tiny tweaks. lgtm otherwise!

Cephian:
OpenAI seems to have gotten their act together and this hardly happens anymore, so I'm just gonna close this as it would probably add needless bloat to the already chaotic LLM module. If it starts happening again we can reconsider.

PR #192, opened 2023-07-31
Change mutagen dependency to git version

Cephian:
We rely on https://github.com/quodlibet/mutagen/pull/601 to be able to detect song lengths for tagless mp3s. However this is not in a current release of Mutagen. The last release of Mutagen was Oct 2022 and a new release seems stalled https://github.com/quodlibet/mutagen/pull/612.

I updated venv packages at some point and reintroduced this "bug" by accident, so I think it's good that `requirements.txt` reflect the real expected requirements. We could additionally choose to pin a specific commit after testing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 discord_emoji>=1.4.1
-mutagen>=1.45.1
+git+https://github.com/quodlibet/mutagen.git
 nextcord[voice]>=2.4.1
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
Good catch. I had forgotten that we were relying unmerged behavior for mutagen. This LGTM.

PR #200, opened 2023-11-07
Add DALL·E 3 cover art generation

### LINKED ISSUE #198: DALL-E Integration (GPT)
Generate album art for songs that do not have art with DALL-E during the 10 second break between songs.

Cephian:
Resolves #198. Generate album art if OpenAI key is available and track has none embedded.

Will time out generation after 20 seconds.

image

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -1,8 +1,11 @@
 import asyncio
 import random
+import time
 from collections import defaultdict
+from io import BytesIO
 from typing import Dict, List, Optional, Tuple, Union
 
+import requests
 from nextcord import (
     Attachment,
     ChannelType,
@@ -10,6 +13,7 @@ from nextcord import (
     ClientException,
     Embed,
     FFmpegPCMAudio,
+    File,
     Interaction,
     Message,
     StageChannel,
@@ -21,6 +25,7 @@ from nextcord import (
 import config
 import discord_utils
 import forms
+import llm
 import persistent_state
 import song_utils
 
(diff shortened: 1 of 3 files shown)


src/bust.py
```diff
@@ -212,20 +217,40 @@ async def finish(self, say_goodbye: bool = True) -> None:
...
-            await self.message_channel.send(embed=embed)
-            await asyncio.sleep(config.seconds_between_songs)
+        # Send the chilling message
+        embed_title = "Currently Chilling"
+        embed_content = (
+            "The track will start soon...\n\n**REMEMBER TO VOTE ON THE GOOGLE FORM!**"
+        )
+        embed = Embed(title=embed_title, description=embed_content)
+        await self.message_channel.send(embed=embed)
+        start_album_generation = time.time()
```
anoadragon453, line 227: Could you add a comment here explaining why we're using `time.time()`?

It's currently not clear just from reading the code.

src/bust.py
```diff
@@ -212,20 +217,40 @@ async def finish(self, say_goodbye: bool = True) -> None:
...
 
         # Pop a song off the front of the queue and play it
         submit_message, attachment, local_filepath = self.bust_content[index]
 
+        # Get cover art
+        cover_art = song_utils.get_cover_art(local_filepath)
+        if cover_art is None and config.openai_api_key:
+            artist, title = song_utils.get_song_metadata(
```
anoadragon453, line 235:
```suggestion
            # Use generative AI to create some album art for this song.
            artist, title = song_utils.get_song_metadata(
```

src/llm.py
```diff
@@ -332,3 +332,32 @@ async def respond(message: Message) -> None:
...
+            model="dall-e-3",
+            prompt=prompt,
+            size="1024x1024",
+            quality="standard",
+            n=1,
+        )
+    except openai.BadRequestError:
+        print(f"Warning: Bad request for prompt {prompt}")
+        return None
+    return response.data[0].url
```
anoadragon453, line 363: Docstrings for these functions please.

src/bust.py
```diff
@@ -212,20 +217,40 @@ async def finish(self, say_goodbye: bool = True) -> None:
...
+            )
+            try:
+                cover_art_url = await asyncio.wait_for(
+                    llm.generate_album_art(artist, title, submit_message.content),
+                    timeout=20.0,
+                )
+                if cover_art_url is not None:
+                    image_data = requests.get(cover_art_url).content
+                    image_bytes_fp = BytesIO(image_data)
+                    cover_art = File(image_bytes_fp, "ai_cover.png")
```
anoadragon453, line 246:
```suggestion
                    # Attaching the URL for the artwork directly to the embed doesn't appear to work,
                    # so instead download the art and upload it to Discord.
                    # This also ensures that the cover art is preserved on Discord's CDN, whereas
                    # the original URL may not remain forever.
                    image_data = requests.get(cover_art_url).content
                    image_bytes_fp = BytesIO(image_data)
                    cover_art = File(image_bytes_fp, "ai_cover.png")
```

src/bust.py
```diff
@@ -212,20 +217,40 @@ async def finish(self, say_goodbye: bool = True) -> None:
...
+            except asyncio.TimeoutError:
+                print("Warning: cover art generation timed out")
+
+        waited = time.time() - start_album_generation
+        time_to_sleep = max(0, config.seconds_between_songs - waited)
+        await asyncio.sleep(time_to_sleep)
```
anoadragon453, line 252: A comment for these lines would also make sense.

src/llm.py
```diff
@@ -332,3 +332,32 @@ async def respond(message: Message) -> None:
...
+async def generate_album_art(
+    artist: str, title: str, description: Optional[str]
+) -> Optional[str]:
+    prompt = [
+        "Generate bizarre photorealistic album art for the following song.",
+        f"{artist} - {title}\n",
+    ]
+    if description:
+        prompt.append(
+            f"Here is how the artist describes the song: {description[:1000]}"
```
anoadragon453, line 346:
```suggestion
            # Cap the description to 1000 characters, to avoid going over any token limits.
            f"Here is how the artist describes the song: {description[:1000]}"
```

src/song_utils.py
```diff
@@ -84,26 +82,44 @@ def song_format(
...
     # We explicitly check for None here, as anything else means that the data was
     # pulled from the audio.
-    if artist is not None:
-        artist = sanitize_tag(artist)
-    if title is not None:
-        title = sanitize_tag(title)
-
-    # Display in the format <Artist-tag> - <Title-tag>
-    # If no artist tag use fallback if valid. Otherwise, skip artist
+    if not artist:
```
anoadragon453, line 85: I don't think we meant to change from `if artist is not None` to `if not artist`, which is different for the case of `artist = ""`.

anoadragon453 (changes requested):
Looks good overall, just a few small changes.

This was fun to do with everyone together!

src/song_utils.py
Cephian: It's not so much that we changed X line to Y and more that we reorganized the entire logic and the first two lines happened to look similar.

In the first version `if artist is not None` was just gating the sanitization of the tag read from the mp3. If it was "" then it would be sanitized and nothing would change. Then a layer `if artist` statement would choose whether the artist was put in the final string

This is closer to a replacement for the `else` clause of the `if artist` later in the original. As it gates whether or not the fallback is used, not sanitization. Then the later `if artist` is the new sanitization gate.

anoadragon453: Right, I hadn't realised this was more than just a typo. Thanks for the detailed explanation!

anoadragon453 (approved):

PR #203, opened 2023-11-14
Add a nix flake to the repo

anoadragon453:
This PR adds a [Nix Flake](https://zero-to-nix.com/concepts/flakes) to the repo, which will install and make available all the dependencies needed to develop the busty bot in a single command:

```
nix develop --impure
```

To use this, [install nix](https://github.com/DeterminateSystems/nix-installer#the-determinate-nix-installer) on your system and run the command above. Python 3.11 will be installed, along with ffmpeg, and a virtualenv will be created at `.venv/`. You'll be dropped into a shell with the [starship prompt](https://starship.rs/) (mostly because I think it looks nice). Exit the shell, and all installed programs will no longer be available - truly isolated developer environments!

This all works thanks to [nix](https://nixos.org/), and works on Linux and MacOS. On Windows, it should work through WSL2.

If you'd like to have this development environment activate on `cd`ing into the `busty` directory, install [direnv](https://direnv.net/) and place a file at `.envrc` with the following contents:

```
use flake . --impure
```

Now the developer environment will automatically activate/deactivate when you cd in/out of the `busty` directory.

---

The `flake.nix` file contains all the configuration, as well as naming some `inputs` for where package and other definitions come from. The `flake.lock` is a lock file that locks the exact git hash of each of those inputs - which ensures that everyone will have the exact same dependencies installed (note: this isn't true for the python deps, until we use something that locks those, like [poetry](https://python-poetry.org/). But we'll all have the same `ffmpeg` version at least).

To update that lock file, one does:

```
nix flake update
```

and commits the new `flake.lock`.

`flake.nix` is written in the nix language. I suggest reading [this article](https://zero-to-nix.com/concepts/nix-language) as a primer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ src/__pycache__/
 
 # nix developer environment
 .devenv/
+.venv
 
 # direnv
 .envrc
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,10 @@ src/__pycache__/
 
 # Bot state, google service keys, llm context
 *.json
+
+# nix developer environment
+.devenv/
+
+# direnv
+.envrc
+.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,235 @@
+{
+  "nodes": {
+    "devenv": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "nix": "nix",
+        "nixpkgs": "nixpkgs",
+        "pre-commit-hooks": "pre-commit-hooks"
+      },
+      "locked": {
+        "lastModified": 1688058187,
+        "narHash": "sha256-ipDcc7qrucpJ0+0eYNlwnE+ISTcq4m03qW+CWUshRXI=",
+        "owner": "cachix",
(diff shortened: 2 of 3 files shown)


Cephian:
Are you planning on migrating the Ubuntu server? Or is this for personal development

anoadragon453:
This is for personal development yeah. The flake would have a package output for production purposes.

I also don't think it's worth managing the VPS with nix as it's already so low effort.

anoadragon453:
The point of making this a PR is just to double-check there's no objections to including it in the repo It also serves as a small introduction to nix development environments, in case you want to give it a try.

I don't expect either of you to learn nix in order to review it, though. So if there are no objections, feel free to :white_check_mark: and move on.

Cephian (approved):
Hmmm Yes I see very interesting. This is excellent quality Nicks code you have here and I would be honored to add it to my collection

anoadragon453:
Thank you very much sir... I am very pleased with this result...
May good tidings come to you and wishes upon your family stump.

PR #204, opened 2023-11-14
Add a missing await

anoadragon453:
Whoopsie.


Cephian:
Oops i accidentally just merged instead of reviewed. Unfortunately this PR was filled with fundamental errors but we will just have to live with it

anoadragon453:
blocked and reported

PR #210, opened 2024-03-12
Allow so-called "mood killer" emoji

Cephian:
imo these are funny and fit busty's personality

also flags was listed twice in the comment about which categories we removed


anoadragon453 (approved):
I think these do make more sense now that we've given Busty a personality. One can imagine they're coming from her, rather than representing the server's general take on a track.

jcogan97 (approved):

PR #211, opened 2024-04-09
Fix unbound bot_state_str variable on first load

Cephian:
Busty crashes due to this unbound variable on line 28 if you have no state file.


anoadragon453 (approved):
Oops. Type-checking ftw.

PR #212, opened 2024-04-10
Show a configurable number of top submitters

### LINKED ISSUE #188: Add top N submitters to /info
Right now we can only see the longest submitter. I'd like to add the ability to see the top N (with probably N=3) just for fun when I get around to it.

Cephian:
Closes #188. This was originally just an idea for fun, but then things got real when we were explicitly asked.
image

Preview image (has also been tested when # of submitters is less than # to be shown)
image

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -386,11 +386,11 @@ class BustController:
 
         # Format list of users with longest total submission length
         submitters_sorted_by_len = sorted(
-            [(len, sub) for (sub, len) in submitter_to_len.items()], reverse=True
+            [(length, sub) for (sub, length) in submitter_to_len.items()], reverse=True
         )
         longest_submitters_formatted = [
-            f"{i+1}. {submitter.mention} - {song_utils.format_time(int(len))}"
-            for i, (len, submitter) in enumerate(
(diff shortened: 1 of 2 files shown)


src/bust.py
```diff
@@ -384,19 +384,26 @@ async def send_stats(self, interaction: Interaction) -> None:
...
-        # User with longest total submission length
-        longest_submitter = max(submitter_to_len, key=submitter_to_len.get)
-        longest_submitter_time = int(submitter_to_len[longest_submitter])
+        # Format list of users with longest total submission length
+        submitters_sorted_by_len = sorted(
+            [(len, sub) for (sub, len) in submitter_to_len.items()], reverse=True
+        )
+        longest_submitters_formatted = [
+            f"{i+1}. {submitter.mention} - {song_utils.format_time(int(len))}"
+            for i, (len, submitter) in enumerate(
```
anoadragon453, line 393: Note that `len` here is over-shadowing the built-in `len`, which may cause some IDEs to spit out warnings. I'd just name it `length`.

anoadragon453 (approved):
LGTM, nice feature!

jcogan97 (approved):
nice one

PR #213, opened 2024-06-06
Update nix flake

Cephian:
I am combining several changes that are probably not worth separating out

- Update flake lock
- Update pinned devenv version
- Add `dev-requirements.txt` in addition to `requirements.txt` to venv
- Switch from python311 to python3
    - Version is pinned by flake lock anyways, actual production is not pinned afaik
- Remove starship
    - For some reason my devenv starship looks like the image below, but removing it from the flake cause my normal system-level starship to work fine. Since it's not really a project dependency it may be best to rely on user shell customization?
image

Changes have been tested

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,7 @@
     # Output a development shell for x86_64/aarch64 Linux/Darwin (MacOS).
     systems.url = "github:nix-systems/default";
     # A development environment manager built on Nix. See https://devenv.sh.
-    devenv.url = "github:cachix/devenv/v1.0.5";
+    devenv.url = "github:cachix/devenv/v1.0.7";
   };
 
   outputs = { self, nixpkgs, devenv, systems, ... } @ inputs:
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -16,7 +16,7 @@
     # Output a development shell for x86_64/aarch64 Linux/Darwin (MacOS).
     systems.url = "github:nix-systems/default";
     # A development environment manager built on Nix. See https://devenv.sh.
-    devenv.url = "github:cachix/devenv/v0.6.3";
+    devenv.url = "github:cachix/devenv/v1.0.5";
```
anoadragon453, line 19: devenv is moving fast lately:

```suggestion
    devenv.url = "github:cachix/devenv/v1.0.7";
```

flake.nix
```diff
@@ -36,11 +36,6 @@
             inherit inputs pkgs;
             modules = [
               {
-                # Make use of the Starship command prompt when this development environment
-                # is manually activated (via `nix develop --impure`).
-                # See https://starship.rs/ for details on the prompt itself.
-                starship.enable = true;
```
anoadragon453, line 42: I realised that if you're using starship (or any other prompt style) as a system prompt, it will continue to be preserved when making use of [direnv](https://github.com/direnv/direnv). So I found in my use case that specifying it at a project level is no longer necessary for me (and just messes with others' prompts).

So happy to remove it.

anoadragon453 (approved):
Thanks!

PR #215, opened 2024-07-17
Add logging module

### LINKED ISSUE #196: Exceptions are not printed when they happen
From previous busty events, Busty will silently stop playing songs entirely and hang on the in-between state inside the queue. We receive nothing from the traceback until Busty is either stopped entirely or restarted.

At restart/shut down, we receive this error from the Traceback.
`application data after close notify (_ssl.c:2702)`

This could be potentially caused by latency between the Busty Server and Discord
[Potential Cause](https://github.com/aio-libs/aiohttp/issues/3535#issuecomment-544279431)

Cephian:
Fixes #196 

Nextcord uses the logging module to log all events. Without the logging module setup, errors and such will just be swallowed. This PR uses recommended settings from https://docs.nextcord.dev/en/stable/logging.html, but to stderr instead of a file.

It's getting a little ugly doing so much setup at a global level, so I plan to do some refactoring of `main.py` in another PR. I also may switch our current print logging to the logging module. But for now I think this as important change to make so we can catch future errors.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1,6 +1,8 @@
 import asyncio
+import logging
 import os
 import random
+import sys
 from typing import Optional
 
 from nextcord import (
@@ -20,6 +22,19 @@ import discord_utils
 import llm
 import persistent_state
 import song_utils
+
+
+def setup_logging(log_level):
+    logger = logging.getLogger("nextcord")
+    logger.setLevel(log_level)
+    handler = logging.StreamHandler(stream=sys.stderr)
+    handler.setFormatter(
+        logging.Formatter("%(asctime)s:%(levelname)s:%(name)s: %(message)s")
+    )
+    logger.addHandler(handler)
+
+
+setup_logging(logging.INFO)
 
 # This is necessary to query guild members
 intents = Intents.default()


src/main.py
```diff
@@ -21,6 +23,19 @@
 import persistent_state
 import song_utils
 
+
+def setup_logging(log_level):
+    logger = logging.getLogger("nextcord")
+    logger.setLevel(log_level)
+    handler = logging.StreamHandler(stream=sys.stderr)
```
jcogan97, line 30:
```suggestion
    handler = logging.FileHandler(stream=sys.stderr, filename='nextcord.log', encoding='utf-8', mode='w')
```

there's some extra parameters that are recommended from the [nextcord docs](https://docs.nextcord.dev/en/stable/logging.html), I don't know if you want to maybe include these as well

jcogan97: LGTM, otherwise

jcogan97: yeah, totally disregard this since the logging on those docs are for writing to a file rather than streaming to a terminal

Cephian: Note in the PR I say this writes to stderr instead of a file, that's what the modification I did from the nextcord example is. I did this so we can just use systemd  to log everything for now since we still have some print logging. Later it would probably be better to move everything to the logging module and maybe then we can revisit whether io streams or a file is better.

Cephian: Oh lol you already replied. cool

anoadragon453, line 30: Streaming all of nextcord's logs to stderr is an interesting choice. Is this to separate the logs from busty's own?

stderr should still be visible in `journalctl` and the command line, so this works for now.

anoadragon453 (approved):
Well that would explain a lot. Thanks for going and figuring this out.

Ideally we'd have the ability to modify the log level from the environment, but happy to save that for a future PR.

Similarly, it would be good to rewrite our own logging to make use of the `logging` module, such that it doesn't look out of place next to the formatted nextcord logs.

src/main.py
Cephian: Yes this PR was intended to be more of a "very quick, for now" type deal just so if it crashes we're ready. Then we accidentally let it sit for two weeks lol

I'll send a follow-up PR with more thought-out organization

PR #216, opened 2025-02-24
Seek to time

JoeCodesThings:
When busty crashes mid-song and the song is rather long, it's often ideal to try to start again in the middle.  This PR adds a parameter to the /bust command called starttime that allows users to enter a time either in seconds (93) or timestamp format (1:15), (01:23:45). Busty will convert the portion of the song into ogg format to avoid chip/pop artifacts introduced by the FFmpegPCMAudio function and play the converted song as Discord intended (because they love ogg sooooo very much), then proceed to the next song as per usual.

This was tested on a Windows machine using NextCord 3.0.1, there were breaking changes to slash_command introduced in 3.0.0 that required me to format the parameters separately in order to run. Let me know and I can revert those if needed.

### DIFF
--- a/src/bust.py
+++ b/src/bust.py
@@ -13,6 +13,7 @@ from nextcord import (
     ClientException,
     Embed,
     FFmpegPCMAudio,
+    FFmpegOpusAudio,
     File,
     Interaction,
     Message,
@@ -28,6 +29,8 @@ import forms
 import llm
 import persistent_state
 import song_utils
+import os
+import subprocess
 
 
 class BustController:
@@ -96,12 +99,13 @@ class BustController:
             self.playing_index = max(0, track_number) - 1
             self.play_song_task.cancel()
 
-    async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
+    async def play(self, interaction: Interaction, skip_count: int = 0, start_time: str = "") -> None:
         """Begin playback.
 
         Args:
             interaction: An interaction which has not yet been responded to.
             skip_count: List index to start playback from.
+            start_time: String in the format of: ##:## specifying start time, or a numeric string in seconds.
         """
 
         await interaction.response.defer(ephemeral=True)
@@ -158,6 +162,9 @@ class BustController:
         await self.message_channel.send("Let's get **BUSTY**.")
         await interaction.delete_original_message()
 
+        # Get first song offset
+        seek_to_seconds = song_utils.convert_timestamp_to_seconds(start_time)
+
         # Play songs
         self.playing_index = skip_count
         while self.playing_index < len(self.bust_content):
@@ -166,7 +173,7 @@ class BustController:
 
             # wrap play_song() in a coroutine so it is cancellable
             self.play_song_task = asyncio.create_task(
-                self.play_song(self.playing_index)
+                self.play_song(self.playing_index, seek_to_seconds)
             )
             try:
                 await self.play_song_task
@@ -174,6 +181,8 @@ class BustController:
                 # Voice client playback must be manually stopped
                 self.voice_client.stop()
 
+            if(seek_to_seconds > 0):
+                seek_to_seconds = 0
             self.playing_index += 1
 
         # tidy up
@@ -216,7 +225,7 @@ class BustController:
         self.original_bot_nickname = None
         self._finished = True
 
-    async def play_song(self, index: int) -> None:
+    async def play_song(self, index: int, timestamp: int) -> None:
         # Send the chilling message
         embed_title = "Currently Chilling"
         embed_content = (
@@ -281,8 +290,12 @@ class BustController:
 
         await discord_utils.try_set_pin(self.now_playing_msg, True)
 
+        temp_filename = f"temp_audio.ogg"
         # Called when song finishes playing
         async def ffmpeg_post_hook(e: Exception = None):
+            if os.path.exists(temp_filename):
+                os.remove(temp_filename)
+                print("Temp song deleted:", temp_filename)
             if e is not None:
                 print("Song playback quit with error:", e)
             # Unpin now playing message
@@ -291,16 +304,39 @@ class BustController:
                 self.now_playing_msg = None
             await play_lock.release()
 
+        # Get optional timestamp
+        song_len = 0
+        timeStr = None
+        if(timestamp > 0):
+            song_len = song_utils.get_song_length(local_filepath)
+            if(timestamp < song_len):
+                timeStr = f" -ss {timestamp}"
+            else:
+                print("Attempted to seek past length of song. Ignoring timestamp.")
+
+
         # Play song
         play_lock = asyncio.Lock()
         # Acquire the lock during playback so that on release, play_song() returns
         await play_lock.acquire()
-        self.voice_client.play(
-            FFmpegPCMAudio(
-                local_filepath, options=f"-filter:a volume={config.VOLUME_MULTIPLIER}"
-            ),
-            after=ffmpeg_post_hook,
-        )
+        if(timeStr is not None):
+            ffmpeg_command = [
+                'ffmpeg', '-i', local_filepath, '-ss', str(timestamp), '-t', str(song_len - timestamp), '-c:a', 'libopus', '-b:a', '128k', '-y', temp_filename
+            ]
+            subprocess.run(ffmpeg_command, check=True)
+
+            self.voice_client.play(
+                FFmpegOpusAudio(
+                    temp_filename, options=f"-filter:a volume={config.VOLUME_MULTIPLIER}"),
+                    after=ffmpeg_post_hook,
+                    )
+        else:
+            self.voice_client.play(
+                FFmpegPCMAudio(
+                    local_filepath, options=f"-filter:a volume={config.VOLUME_MULTIPLIER}"),
+                    after=ffmpeg_post_hook,
+                    )
+        
 
         # Set now playing title
         self.now_playing_str = song_utils.song_format(
--- a/src/bust.py
+++ b/src/bust.py
@@ -74,6 +74,7 @@ class BustController:
 
         self._finished: bool = False
         self._playing_index: Optional[int] = None
+        self.temp_audio_file: str = f"temp_audio.ogg"
 
     def is_active(self) -> bool:
         return self.voice_client and self.voice_client.is_connected()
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -354,6 +360,16 @@ async def on_application_command_error(
     else:
         print(error)
 
+# NextCord 3+ requires dm_permission parameters outside of the slash_command declaration
```
Cephian, line 363: Can you split the slash commands thing into a separate PR? Also, I'm only skimming but it seems to me like the proper way to do this might be through the `contexts` field from `slash_command` now. The default `IntegrationContextType` allows all contexts, and we want specifically the guild context.

slash_command: https://docs.nextcord.dev/en/stable/api.html#nextcord.Client.slash_command
IntegrationContextType: https://docs.nextcord.dev/en/stable/api.html#nextcord.InteractionContextType

src/song_utils.py
```diff
@@ -232,3 +232,40 @@ def get_cover_art(filename: str) -> Optional[File]:
...
+
+def convert_timestamp_to_seconds(time_str):
+    # Converts a time string into seconds. Returns 0 if format is invalid.
+    # Format is handled either in pure seconds (93, 180) or hh:mm:ss format (1:23:45).
+    if time_str is None:
+        return 0
+    
+    if ':' not in time_str:
+        if not time_str.isdigit():
+            return 0
```
Cephian, line 244: imo preferred behavior would be to return `None` on all invalid inputs, and then in the bust command instead of ignoring the parameter, just respond to the interaction refusing to bust bc of malformed input. If someone is using the input, they probably mean to and did not just make a typo and so an error would be better so they can try again, instead of just ignoring their input

src/bust.py
```diff
@@ -295,12 +322,21 @@ async def ffmpeg_post_hook(e: Exception = None):
...
-                local_filepath, options=f"-filter:a volume={config.VOLUME_MULTIPLIER}"
-            ),
-            after=ffmpeg_post_hook,
-        )
+        if(timestamp > 0):
+            self.seek_and_convert_to_opus(timestamp, local_filepath)
+
+            self.voice_client.play(
+                FFmpegOpusAudio(
+                    self.temp_audio_file, options=f"-filter:a volume={config.VOLUME_MULTIPLIER}"),
```
Cephian, line 330: I think we should be able to keep the actual `self.voice_client.play` call with all its options outside the if statement (avoiding duplication) if we do something like

```python
file_to_play = local_filepath
if (timestamp > 0):
    self.seek_and_convert_to_opus(timestamp, local_filepath)
    file_to_play = self.temp_audio_file
```
and then put the voice_client.play after

src/bust.py
```diff
@@ -166,14 +190,16 @@ async def play(self, interaction: Interaction, skip_count: int = 0) -> None:
...
-                self.play_song(self.playing_index)
+                self.play_song(self.playing_index, seek_to_seconds)
             )
             try:
                 await self.play_song_task
             except asyncio.CancelledError:
                 # Voice client playback must be manually stopped
                 self.voice_client.stop()
 
+            if(seek_to_seconds > 0):
```
Cephian, line 201: seek to seconds will always be nonnegative right? so I think we can just remove the if statement and put
`seek_to_seconds = 0`
worst case it's just a single nullop assignment

src/bust.py
```diff
@@ -88,6 +92,22 @@ def stop(self) -> None:
...
+            os.remove(self.temp_audio_file)
+
+    def seek_and_convert_to_opus(self, timestamp: int, local_filepath: str) -> None:
+        song_len = song_utils.get_song_length(local_filepath)
+        if(timestamp >= song_len):
+            timestamp = 0
+            print("Attempted to seek past length of song. Ignoring timestamp.")
+
+        ffmpeg_command = [
+            'ffmpeg', '-i', local_filepath, '-ss', str(timestamp), '-t', str(song_len - timestamp), '-c:a', 'libopus', '-b:a', '128k', '-y', self.temp_audio_file
```
Cephian, line 108: I haven't tried this, but I think the `-t` flag part isn't needed? I think it should go from the `-ss` time to the end automatically, though I have not tested this and can do so later. Setting an end time like this feels like it could pootentially chop some samples or so off the end in some cases due to imprecision

src/bust.py
```diff
@@ -71,6 +74,7 @@ def __init__(
 
         self._finished: bool = False
         self._playing_index: Optional[int] = None
+        self.temp_audio_file: str = f"temp_audio.ogg"
```
Cephian, line 77: Busty is technically multiserver (untested lol) and having a single temp audio file name could cause conflicts between multiple busts happening on different servers. I think with this implementation the temp file will also be dumped into the central busty repo folder instead of being contained. See code linked below for where and how we save attachment files, and maybe make the temp audio file something like `attachments/<guild_id>/temp_audio.ogg` instead

https://github.com/anoadragon453/busty/blob/main/src/discord_utils.py#L37-L40

src/bust.py
```diff
@@ -13,6 +13,7 @@
     ClientException,
     Embed,
     FFmpegPCMAudio,
+    FFmpegOpusAudio,
```
Cephian, line 16: I think this import is unsorted, run scripts-dev/lint.sh to sort, format, and run linters that we use for github automation

Cephian (changes requested):
haven't run any code, this is just from skimming it in the browser. thanks for the PR!

src/bust.py
JoeCodesThings: This was done to make sure busty doesn't seek to time all concurrent audio files - arguably I'm not sure if this is the ideal spot to put it though

src/main.py
JoeCodesThings: Will do

src/song_utils.py
```diff
@@ -232,3 +232,41 @@ def get_cover_art(filename: str) -> Optional[File]:
 
     # Create a new discord file from the file pointer and name
     return File(image_bytes_fp, filename=cover_filename)
+
+
+def convert_timestamp_to_seconds(time_str: str):
+    # Converts a time string into seconds. Returns 0 if format is invalid.
```
Cephian, line 238: I think this returns `None` if format is invalid, not 0

src/song_utils.py
```diff
@@ -232,3 +232,41 @@ def get_cover_art(filename: str) -> Optional[File]:
...
+    # Format is handled either in pure seconds (93, 180) or hh:mm:ss format (1:23:45).
+    if time_str is None:
+        return None
+
+    if ":" not in time_str:
```
Cephian, line 243: Nit: You can remove this check and it should still have the same behavior (less efficient by a tiny amt but totally inconsequential, I think clarity is more valuable here)

Cephian (commented):
On my first test I get the error dumped below. On the frontend it just manifested as playing the song for a little while after I seeked (I assume doing the ffmpeg cutting in the bg) and then just playing the next song

```
2025-05-04 23:18:57,888:INFO:nextcord.voice_client: Connecting to voice...
2025-05-04 23:18:57,888:INFO:nextcord.voice_client: Starting voice handshake... (connection attempt 1)
2025-05-04 23:18:58,053:INFO:nextcord.voice_client: Voice handshake complete. Endpoint found santa-clara2975.discord.media
2025-05-04 23:18:58,196:INFO:nextcord.gateway: selected the voice protocol for use (xsalsa20_poly1305_lite)
2025-05-04 23:18:58,213:INFO:nextcord.gateway: received secret key for voice connection
2025-05-04 23:19:14,302:INFO:nextcord.player: Preparing to terminate ffmpeg process 948007.
2025-05-04 23:19:14,303:INFO:nextcord.player: ffmpeg process 948007 has not terminated. Waiting to terminate...
2025-05-04 23:19:14,306:INFO:nextcord.player: ffmpeg process 948007 should have terminated with a return code of -9.
[in#0 @ 0x490cc0] Error opening input: No such file or directory
Error opening input file attachments/765769500053864468/temp_audio.ogg.
Error opening input files: No such file or directory
2025-05-04 23:19:16,692:INFO:nextcord.player: Preparing to terminate ffmpeg process 948099.
2025-05-04 23:19:16,692:INFO:nextcord.player: ffmpeg process 948099 successfully terminated with return code of 254.
```

src/song_utils.py
Cephian: by "this check" i mean at line 243, idk why it highlighted the wrong section

src/bust.py
```diff
@@ -295,10 +353,22 @@ async def ffmpeg_post_hook(e: Exception = None):
...
+        if self._seek_to_seconds is not None:
+            audio_to_play = FFmpegOpusAudio(
+                self.temp_audio_file,
+                options=f"-filter:a volume={config.VOLUME_MULTIPLIER}",
+            )
+            self._seek_to_seconds = None
+        else:
+            audio_to_play = FFmpegPCMAudio(
+                local_filepath,
+                options=f"-filter:a volume={config.VOLUME_MULTIPLIER}",
```
Cephian, line 367: Actually, there's something I'm curious about. I noticed we can pass in arbitrary ffmpeg options when playing audio. Instead of cutting the audio with ffmpeg and then reloading the cut audio in separate steps, can't you just pass in the `            "-ss <timestamp>` directly into the options when playing? This would eliminate all the system IO and double ffmpeg pass overhead

JoeCodesThings:
> On my first test I get the error dumped below. On the frontend it just manifested as playing the song for a little while after I seeked (I assume doing the ffmpeg cutting in the bg) and then just playing the next song
> 
> ```
> 2025-05-04 23:18:57,888:INFO:nextcord.voice_client: Connecting to voice...
> 2025-05-04 23:18:57,888:INFO:nextcord.voice_client: Starting voice handshake... (connection attempt 1)
> 2025-05-04 23:18:58,053:INFO:nextcord.voice_client: Voice handshake complete. Endpoint found santa-clara2975.discord.media
> 2025-05-04 23:18:58,196:INFO:nextcord.gateway: selected the voice protocol for use (xsalsa20_poly1305_lite)
> 2025-05-04 23:18:58,213:INFO:nextcord.gateway: received secret key for voice connection
> 2025-05-04 23:19:14,302:INFO:nextcord.player: Preparing to terminate ffmpeg process 948007.
> 2025-05-04 23:19:14,303:INFO:nextcord.player: ffmpeg process 948007 has not terminated. Waiting to terminate...
> 2025-05-04 23:19:14,306:INFO:nextcord.player: ffmpeg process 948007 should have terminated with a return code of -9.
> [in#0 @ 0x490cc0] Error opening input: No such file or directory
> Error opening input file attachments/765769500053864468/temp_audio.ogg.
> Error opening input files: No such file or directory
> 2025-05-04 23:19:16,692:INFO:nextcord.player: Preparing to terminate ffmpeg process 948099.
> 2025-05-04 23:19:16,692:INFO:nextcord.player: ffmpeg process 948099 successfully terminated with return code of 254.
> ```

Interesting, so it tried to play it before it finished converting which wasn't supposed to happen. Did it play where you seeked in the first song?

Cephian:
Neither, it just skipped to the next song.

src/bust.py
Cephian: I see, in that case we can go with this for now and look into faster alternatives later if it becomes a pain point. This feels like something that should be somehow technically possible but might be a pain to figure out in practice.

JoeCodesThings:
> Neither, it just skipped to the next song.

Gotcha. I think I found the issue. The post-hook deletes the temp audio after each song plays so it must have deleted it at the wrong time. I changed it to only delete at the end of a bust or on stop command. FFmpeg will happily overwrite this file during subsequent seeks.

Cephian:
imo no need to handle deletion at all, busty will clear all unreferred to cached items in the attachment download folder on the next list

src/song_utils.py
Cephian: FYI you should mark comment threads as resolved when you fix them so that I can see that they're concluded instead of waiting for my input on something

PR #218, opened 2025-02-25
NextCord 3+

JoeCodesThings:
This PR allows busty to compile and run on NextCord 3.0.0+.


Cephian:
Can you update nextcord min version in requirements.txt?

PR #219, opened 2025-05-03
Pass after when scraping to work around bug in nextcord

Cephian:
There are some opinionated changes here so feel free to add feedback or alternative suggestions.

I thought busty was in need of a bit of a revamp, since it often feels pretty non-ergonomic when I have to go in and mess with stuff, set up venvs on new computers, etc. This PR kept growing in scope (oops), it does the following:

## Changes
1) Change packaging system to [poetry](https://python-poetry.org/)
2) Make busty into a package of her own. This causes internal imports to go from `import XYZ` to `import busty.XYZ`, and also moves all source to `src/busty`.
3) Update flake and flake dependencies.
4) Replace `flake8`, `isort`, and `black` with the more modern [ruff](https://github.com/astral-sh/ruff). Also remove the lint check script as it's all replaced by `ruff`.
5) Update nix flake locked version.
6) Remove types-pillow as the Pillow package includes type annotations or type stubs since version 10.3.0.
7) Updated README with poetry changes

I plan on also consolidating configs between the `llm_context.json` and env var config into something more reasonable like toml/yaml, and moving attachments to a more reasonable cache dir that's not the source folder. And probably revamping and actually documenting the LLM stuff too.


anoadragon453:
Taking off my review request while this is marked as draft.

